PR #3, opened 2016-06-30
Adding extensions sonic words

FranDepascuali:
## Summary ##
**Trello card**: https://trello.com/c/nXWYQKOT/35-extraer-extensiones-utilities-de-sonicwords-a-wolmo-core

Adds extensions and utilities used in SonicWord


woloxci:

Core/Extensions/Foundation/NSBundle.swift
```diff
@@ -10,8 +10,17 @@ import Foundation
...
 
-    public func loadNib(name: String) -> AnyObject {
-        return self.loadNibNamed(name, owner: self, options: nil)[0]
+    public func loadNib<T>(name: String) -> T {
+        // swiftlint:disable force_cast
+        return loadNibNamed(name, owner: self, options: nil)[0] as! T
+        // swiftlint:enable force_cast
+    }
+    
+    public func get(key: String) -> String {
```
guidomb: What about using `subscript` for this?

guidomb:
@FranDepascuali Let me now what you think about using `subscript` instead of `get` in the `NSBundle` extension. Other than that I think this is ready to be merged.

Core/Extensions/Foundation/NSBundle.swift
```diff
@@ -10,8 +10,17 @@ import Foundation
 
 public extension NSBundle {
 
-    public func loadNib(name: String) -> AnyObject {
-        return self.loadNibNamed(name, owner: self, options: nil)[0]
+    public func loadNib<T>(name: String) -> T {
+        // swiftlint:disable force_cast
+        return loadNibNamed(name, owner: self, options: nil)[0] as! T
```
FranDepascuali: @guidomb What do you think of this `as! T`. Do you think we should explictly unwrap it or return a `T?`?

Core/Extensions/Foundation/NSBundle.swift
```diff
@@ -10,8 +10,17 @@ import Foundation
...
+        // swiftlint:enable force_cast
+    }
+    
+    public func get(key: String) -> String {
+        guard let value = objectForInfoDictionaryKey(key) as? String where !value.isEmpty else {
+            fatalError("Key: \"\(key)\" has not been configured in Info.plist")
```
FranDepascuali: @guidomb What do you think of this fatalError? Should we return a `String?`?

Core/Extensions/Foundation/NSBundle.swift
guidomb: Lets make it optional. This would force to do the explicitly unwrap when loading the view but I think it would be better because is more explicit.

Core/Extensions/Foundation/NSBundle.swift
guidomb: I guess that if we want to follow the latest swift conventions we should go with the safest choice. The thing is that we don't use `NSBundle` that often and only to load view so if there is something wrong is pretty easy to catch but I guess that is always good to have some extra safety. That being said, when you are using this API probably there isn't much you can do if this `nil`. So the caller would probably explicit unwrap it or use guard. But I guess is better to make it explicitly in the caller's side that to hide it with a fatal exception.

Core/Extensions/Foundation/NSBundle.swift
```diff
@@ -9,9 +9,22 @@
...
-    public func loadNib(name: String) -> AnyObject {
-        return self.loadNibNamed(name, owner: self, options: nil)[0]
+    
+    public func loadNib<NibType, T: RawRepresentable where T.RawValue == String>(nibName: T) -> NibType {
+        return loadNib(nibName.rawValue)
+    }
+    
+    public func loadNib<NibType>(nibName: String) -> NibType {
+        // swiftlint:disable force_cast
+        return loadNibNamed(nibName, owner: self, options: nil)[0] as! NibType
```
FranDepascuali: @guidomb what do you think of the force cast `as! NibType`? Should we change it to `as?` and left the caller decide what to do?

Core/Extensions/Foundation/NSBundle.swift
```diff
@@ -9,9 +9,22 @@
...
+        // swiftlint:enable force_cast
+    }
+    
+    public subscript(key: String) -> String {
+        guard let value = objectForInfoDictionaryKey(key) as? String where !value.isEmpty else {
+            fatalError("Key: \"\(key)\" has not been configured in Info.plist")
```
FranDepascuali: @guidomb what do you think of the `fatalError`? Should we return .None and leave the caller decide what to do?

woloxci:

Core/Extensions/Foundation/NSBundle.swift
guidomb: Read my comment above. Make the `subscript` return a `String?`

Core/Extensions/Foundation/NSBundle.swift
guidomb: Use `as?` and change the type signature to `NibType?`

guidomb:
:+1:

Core/Extensions/AVFoundation/AVAsset.swift
```diff
@@ -0,0 +1,38 @@
...
+                var error: NSError?
+                
+                for key in keys {
+                    let status = self.statusOfValueForKey(key, error: &error)
+                    // We need to ensure that the keys were loaded
+                    guard status == .Loaded else {
+                        if let error = error {
+                            // There is an error only if the status is Failed
+                            observer.sendFailed(error)
+                        }
```
danielaRiesgo: Shouldn't we send a Completed to the observer or something before returning? Or are we sure it will always enter the if body?

Core/Extensions/Foundation/String.swift
```diff
@@ -8,7 +8,7 @@
 
 import Foundation
 
-public func/(lhs: String, rhs: String) -> String {
+public func / (lhs: String, rhs: String) -> String {
```
danielaRiesgo: In the NSURL extension we didn't use spaces. For me it's the same which one we choose, as long as we use the same in all.

Core/Extensions/ReactiveCocoa/Signal.swift
```diff
@@ -0,0 +1,22 @@
...
+
+import ReactiveCocoa
+
+public extension Signal {
+    
+    /// Transforms a `Signal<Value, Error>` to `SignalProducer<Value, NewError>`
+    /// This is usually pretty useful when the `flatMap` operator is used and the outer
+    /// producer as `NoError` and the inner one a different type of error.
+    ///
+    /// - returns: A signal producer with the same value type but with `NewError` as the error type
```
danielaRiesgo: `- returns: A signal with the same ...`

woloxci:

Core/Extensions/UIKit/UIView.swift
```diff
@@ -70,16 +70,87 @@ extension UIView {
...
+/**
+ Uncollapse a view by removing/modifying constraint height.
+ 
+ **Warning:** It is intended to be used to uncollapse a view *without* subviews.
+ 
+ - Parameter view: The view to collapse
+ - Parameter animated: Indicates if the collapse should be animated.
+ - Parameter animationDuration: The animationDuration of the collapse.
+ */
+func uncollapse(view: UIView, animated: Bool = true, animationDuration: NSTimeInterval = 1) {
```
danielaRiesgo: Add `public` access level.

Core/Extensions/UIKit/UIView.swift
```diff
@@ -70,16 +70,87 @@ extension UIView {
...
+        // If we have a previous height, it means that before collapsing it had another height constraint.
+        if let previousHeight: CGFloat = getAssociatedObject(view, key: &hideableUIViewConstraintKey) {
+            heightConstraint.constant = previousHeight
+        } else {
+            // We desactive the height constraint added because it didn't have a height before collapsing.
+            heightConstraint.active = false
+        }
+    }
+    
+    view.setNeedsLayout()
```
danielaRiesgo: This is repeated in both functions. Maybe we could extract it to another private function.

Core/Extensions/UIKit/UIView.swift
```diff
@@ -70,16 +70,87 @@ extension UIView {
...
+/**
+ Collapse a view by adding/modifying constraint height.
+ 
+ - Warning: It is intended to be used to collapse a view *without* subviews.
+ 
+ - Parameter view: The view to collapse
+ - Parameter animated: Indicates if the collapse should be animated.
+ - Parameter animationDuration: The animationDuration of the collapse.
+ */
+public func collapse(view: UIView, animated: Bool = true, animationDuration: NSTimeInterval = 1) {
```
danielaRiesgo: I think these functions (collapse and uncollapse) could be an extension of UIView.

Core/UIAlertController.swift
```diff
@@ -0,0 +1,46 @@
+//
```
danielaRiesgo: Do you think this file should be here or at Utilities? I don't know if it should go in Utilities or not, but I don't like it not being inside a folder which indicates its use or category.

Core/UIAlertController.swift
```diff
@@ -0,0 +1,46 @@
...
+        
+        return alertController
+    }
+    
+    public static func errorAlertController(alertViewModel: ErrorAlertViewModel) -> UIAlertController {
+        let alertController = UIAlertController(title: alertViewModel.title,
+                                                message: alertViewModel.message,
+                                                preferredStyle: UIAlertControllerStyle.Alert)
+        
+        let dismissAction = UIAlertAction(title: alertViewModel.dismissButtonTitle, style: UIAlertActionStyle.Default) { _ in
```
danielaRiesgo: Maybe the style can be in the `ErrorAlertViewModel` as well.

Core/UIAlertController.swift
```diff
@@ -0,0 +1,46 @@
...
+import UIKit
+
+public extension UIAlertController {
+    
+    public static func confirmationAlertController(alertViewModel: ConfirmationAlertViewModel) -> UIAlertController {
+        let alertController = UIAlertController(title: alertViewModel.title,
+                                                message: alertViewModel.message,
+                                                preferredStyle: UIAlertControllerStyle.Alert)
+        
+        let confirmAction = UIAlertAction(title: alertViewModel.confirmButtonTitle, style: UIAlertActionStyle.Default) { _ in
```
danielaRiesgo: Maybe the style can be in the `ConfirmationAlertViewModel` as well.

Core/UIAlertController.swift
```diff
@@ -0,0 +1,46 @@
...
+            alertViewModel.confirmAction(alertViewModel)
+        }
+        
+        let dismissAction = UIAlertAction(title: alertViewModel.dismissButtonTitle, style: UIAlertActionStyle.Default) { _ in
```
danielaRiesgo: Maybe the style can be in the `ConfirmationAlertViewModel` as well.

Core/Utilities/Alerts/AlertPresenter.swift
```diff
@@ -0,0 +1,31 @@
...
+    
+    func presentAlert(alertViewModel: ErrorAlertViewModel, completion: (() -> Void)?)
+    
+    func presentAlert(alertViewModel: ConfirmationAlertViewModel, completion: (() -> Void)?)
+    
+}
+
+extension UIViewController: AlertViewPresenter {
+    
+    public func presentAlert(alertViewModel: ErrorAlertViewModel, completion: (() -> Void)? = .None) {
```
danielaRiesgo: I would add a comment indicating that completion will be called after the alert appears in the screen, and not after it disappears (for that use the alert view model's dismiss/confirm action).

Core/Utilities/Alerts/AlertPresenter.swift
```diff
@@ -0,0 +1,31 @@
...
+        let alert = UIAlertController.errorAlertController(alertViewModel)
+        presentViewController(alert, animated: true, completion: completion)
+    }
+    
+    public func presentAlert(alertViewModel: ConfirmationAlertViewModel, completion: (() -> Void)? = .None) {
```
danielaRiesgo: I would add a comment indicating that completion will be called after the alert appears in the screen, and not after it disappears (for that use the alert view model's dismiss/confirm action).

Core/Utilities/Alerts/ConfirmationAlertViewModel.swift
```diff
@@ -0,0 +1,51 @@
...
+    /// - parameter title: The alert title.
+    /// - parameter message: The alert message.
+    /// - parameter dismissButtonTitle: The dismiss button title.
+    /// - parameter dismissAction: The dismiss button action.
+    /// - parameter confirmButtonTitle: The confirm button title.
+    /// - parameter confirmAction: The confirm button action.
+    public init(
+        title: String,
+        message: String,
+        dismissButtonTitle: String = "confirmation-alert-view.dismiss.title".localized,
```
danielaRiesgo: I don't think I would use a localized string for this, because if this would be a framework people using wouldn't realize what string they should localize. Maybe we could add an static property `DefaultDismissButtonTitle` ?

Core/Utilities/Alerts/ConfirmationAlertViewModel.swift
```diff
@@ -0,0 +1,51 @@
...
+        dismissAction: ConfirmationAlertViewModel -> () = { _ in },
+        confirmButtonTitle: String = "confirmation-alert-view.confirm.title".localized,
```
danielaRiesgo: Same as with Confirmation's `dismissButtonTitle`

Core/Utilities/Alerts/ErrorAlertViewModel.swift
```diff
@@ -0,0 +1,27 @@
...
+public struct ErrorAlertViewModel {
+    public let title: String
+    public let message: String
+    public let dismissButtonTitle: String
+    public let dismissAction: ErrorAlertViewModel -> ()
+    
+    public init(
+        title: String,
+        message: String,
+        dismissButtonTitle: String = "error-alert-view.dismiss.title".localized.uppercaseString,
```
danielaRiesgo: Same as with Confirmation's `dismissButtonTitle`.

Core/Utilities/ProgressHUD/ProgressHUDDelegate.swift
```diff
@@ -0,0 +1,37 @@
...
+    
+    func showProgressHud()
+    
+    func showProgressHud(string: String)
+    
+    func hideProgressHud()
+    
+}
+
+extension UIViewController: ProgressHUDDelegate {
```
danielaRiesgo: I would add a comment indicating that these freezes the UI. (Or at least that is what I remember happens. If not, I would a comment saying that it doesn't)

Core/Utilities/ProgressHUD/ProgressHUDDelegate.swift
```diff
@@ -0,0 +1,37 @@
...
+    
+    public func showProgressHud() {
```
danielaRiesgo: I would add a comment indicating that these freezes the UI. (Or at least that is what I remember happens. If not, I would a comment saying that it doesn't)

Core/Utilities/ProgressHUD/ProgressHUDDelegate.swift
```diff
@@ -0,0 +1,37 @@
...
+        MBProgressHUD.showHUDAddedTo(view, animated: true)
+    }
+    
+    public func showProgressHud(string: String) {
```
danielaRiesgo: I would add a comment indicating that these freezes the UI. (Or at least that is what I remember happens. If not, I would a comment saying that it doesn't)

Core/Extensions/AVFoundation/AVAsset.swift
guidomb: Nope. Failure is a termination event.

Core/Extensions/UIKit/UIView.swift
guidomb: Good idea!

Core/UIAlertController.swift
guidomb: I think is this case is OK because the method name says `AlertController` so it implies that is going to create an alert.

Core/UIAlertController.swift
guidomb: Same as before

Core/UIAlertController.swift
guidomb: Same as before

Core/Extensions/AVFoundation/AVAsset.swift
danielaRiesgo: But are we sure it will always enter the if body? Or should we do `else { observer.sendCompleted() }` ?

Core/Utilities/Alerts/ConfirmationAlertViewModel.swift
guidomb: mmm I don't know about that. Is a default parameter if you want to change you can either define that key or set a new value to that parameter

Core/Utilities/Alerts/ConfirmationAlertViewModel.swift
guidomb: Same as before

Core/Extensions/AVFoundation/AVAsset.swift
guidomb: Ohh that is true. I don't what are the other possible value for status but yet we should terminate the observer in that case. The other question is should we send a void element o just completed?

Core/UIAlertController.swift
danielaRiesgo: Sorry, I mean in the `UIAlertActionStyle.Default` of the dismissAction.

guidomb: OK that one is a little more subjective I guess. The only valid options I think could be used here are `Default` or `Destructive`. Maybe the `ErrorAlertViewModel` could have a property with its one `DismissButtonStyle` enum and then do the translation here.

Core/Utilities/Alerts/ConfirmationAlertViewModel.swift
danielaRiesgo: But maybe you want all confirmation alerts' dismiss buttons to have the same title and so you want that to be the default. If this is a framework you download with carthage. The only way you can get the name of the string you have to localize is to read the source or detailed documentation (and so we should add on top what the default value is). If it is a property or something you can see it when visiting the class.

guidomb: Personally I think that if you want all alerts to have the same dismiss title then this is a good approach because it forces the user to think about i18n. This should clearly be detailed in the method's documentation. The alternative would be to have a public mutable class variable. In that case I would go with the i18n approach.

What do you suggest if not?

danielaRiesgo: But maybe the property could be `DismissButtonTitleToLocalize` or something like that, that defaults to `"confirmation-alert-view.dismiss.title"`. I agree that it's good to require i18n, but I just don't like that the string which you have to copy in the strings file and put a value to is "hidden".

guidomb: I agree with the "hidden" part. We could extract this to static immutable constant inside the class.

Core/Extensions/Foundation/String.swift
FranDepascuali: I prefer leaving it with a space. As they are functions, the style rules that apply for functions should apply to operators. What do you think @danielaRiesgo @guidomb ?

guidomb: I would do whatever is doing Apple for the standard library. Maybe there is a rule in the Swift Style guide.

FranDepascuali: https://github.com/github/swift-style-guide
I'll leave it with a space, this convinced me: 
```Rationale: Operators consist of punctuation characters, which can make them difficult to read when immediately followed by the punctuation for a type or value parameter list. Adding whitespace separates the two more clearly.
```

Core/Extensions/AVFoundation/AVAsset.swift
```diff
@@ -10,27 +10,29 @@ import AVFoundation
...
-    public func loadValuesAsynchronouslyForKeys(keys: [String]) -> SignalProducer<(), NSError> {
+    public func loadValuesAsynchronouslyForKeys(keys: [String]) -> SignalProducer<[String: AVKeyValueStatus], NSError> {
         return SignalProducer { observer, _ in
             self.loadValuesAsynchronouslyForKeys(keys) { _ in
                 
-                var error: NSError?
+                var keysStatus: [String: AVKeyValueStatus] = [:]
                 
-                for key in keys {
+                keys.forEach { key in
```
guidomb: You need to change this back to a for enumeration because the return statement inside the closure is exiting the closure not breaking the loop and exiting the function.

guidomb:
@FranDepascuali I've made a comment in an outdated diff but the patch remains in the other commits. Check the `AVAsset` extension, you need to replace the `.forEach` invocation for a `for-in` loop.

Core/Extensions/AVFoundation/AVAsset.swift
```diff
@@ -0,0 +1,40 @@
...
+                    if status == .Failed, let error = error {
+                        observer.sendFailed(error)
+                        return
+                    }
+                    
+                    keysStatus[key] = status
+                }
+                
+                observer.sendNext(keysStatus)
+                observer.sendCompleted()
```
FranDepascuali: @guidomb @danielaRiesgo I don't know if this is the best thing to do. Right now, if there is an error in a key, it will send the error and return, meaning that it is not reporting all the errors. And also, maybe a key status is .Failed and it should not be an error (because other keys were loaded correctly and that particular key could .failed but it is still considered succesfull). And last, we are just returning an error and maybe it should be a couple key, error or a dictionary of all the keys that failed. 
The question arises on the definition of the function, so how should the SignalProducer that it returns be parametrized?

woloxci:

Core/Extensions/UIKit/UIView/Collapsable.swift
```diff
@@ -0,0 +1,88 @@
...
+            UIView.animateWithDuration(animationDuration) {
+                self.layoutIfNeeded()
+            }
+        } else {
+            layoutIfNeeded()
+        }
+    }
+}
+
+private var CollapsableHeightConstraintKey: UInt8 = 0
```
danielaRiesgo: I would put this before the protocol implementation, where it is used.

Core/Extensions/UIKit/UIView/UIView.swift
```diff
@@ -0,0 +1,78 @@
+//
+//  UIView.swift
```
danielaRiesgo: I would part this in two files and name them according to the functionality: like Bordering and ContainerView or something like that.

Core/Utilities/Alerts/ConfirmationAlertViewModel.swift
```diff
@@ -48,4 +52,15 @@ public struct ConfirmationAlertViewModel {
         self.confirmButtonTitle = confirmButtonTitle
     }
     
+    /**
+     Default title for confirm button. You can provide a different default title by changing this property.
+     - seealso: Localizable.String
+     */
+    public static var DefaultConfirmButtonTitle: String = "confirmation-alert-view.dismiss.title".localized.capitalizedString
```
danielaRiesgo: With @guidomb we thought of making it an immutable property. Or at least move the localized part to the default parameter, so that localization is required. What do you say?

Core/Extensions/UIKit/UIView/Collapsable.swift
```diff
@@ -0,0 +1,88 @@
...
+
+public protocol Collapsable {
+    
+    func collapse(animated: Bool, animationDuration: NSTimeInterval)
+    
+    func uncollapse(animated: Bool, animationDuration: NSTimeInterval)
+    
+}
+
+extension UIView: Collapsable {
```
danielaRiesgo: Do we want this to be  `extension UIView: Collapsable` or  `extension Collapsable where Self: UIView` ? Opinions?

Core/Extensions/UIKit/UIView/NibLoadable.swift
```diff
@@ -0,0 +1,29 @@
...
+
+import Foundation
+
+public protocol NibLoadable {
+    
+    static func loadFromNib<T: UIView>(bundle: NSBundle) -> T
+    
+}
+
+extension UIView: NibLoadable {
```
danielaRiesgo: Do we want this to be `extension UIView: NibLoadable` or `extension NibLoadable where Self: UIView` ? Opinions?
In this case I prefer the second option because for sure nor all UIViews will be NibLoadable, only the ones we create a xib for, so I think it's good to require the user to specify the view is NibLoadable.

Core/Utilities/ProgressHUD/ProgressHUDDelegate.swift
```diff
@@ -11,23 +11,25 @@ import MBProgressHUD
...
-    func showProgressHud()
-    
-    func showProgressHud(string: String)
+    func showProgressHud(labelText: String?)
     
     func hideProgressHud()
     
 }
 
 extension UIViewController: ProgressHUDDelegate {
```
danielaRiesgo: Do we want this to be `extension UIViewController: ProgressHUDDelegate` or `extension ProgressHUDDelegate where Self: UIViewController` ? Opinions?

Core/Extensions/AVFoundation/AVAsset.swift
guidomb: If the whole function can fail then the producer should fail. If the invocation of the function has "successfull" then I think the best is to return a `[String : Result<AVKeyValueStatus, NSError>]` or something like that.

Core/Utilities/Alerts/ConfirmationAlertViewModel.swift
guidomb: Yeah I would make it immutable. If you want to change this default value has add the key to your i18n file.

FranDepascuali: I don't like to force the user to add a key to their localizable. However, if you both prefer that approach, I'll do it that way

Core/Extensions/AVFoundation/AVAsset.swift
danielaRiesgo: I think the purpose of this function when called is to get notified when all the keys specified have been loaded. So it's weird to not fail if everything isn't .Loaded. But it can be . Loading, so it may be successful afterwards, so maybe in that case it shouldn't fail and return that dictionary. But if anything is .Failed or .Cancelled I think the whole should fail. If it is .Unknown, I don't know, I guess it should fail too because it's not like it is Loading.
But if the purpose is just to try to load and get notified of the result, then I guess it shouldn't fail and it could send what @guidomb says.

Core/Extensions/UIKit/UIView/Collapsable.swift
FranDepascuali: I put it there because I followed our design guidelines in leaving what is public first and then private things.

Core/Extensions/AVFoundation/AVAsset.swift
guidomb: Personally I don't think that the producer should fail if one of the key has failed. Another option would be to send a next event per key and first send all the successfully loaded keys and the send a failure event for the first key that failed.

guidomb:
👍

danielaRiesgo:
👍

Core/Utilities/AssociatedObject.swift
```diff
@@ -0,0 +1,37 @@
...
+    init(_ value: T) {
+        self.value = value
+    }
+}
+
+private func lift<T>(value: T) -> AssociatedObjectBox<T> {
+    return AssociatedObjectBox(value)
+}
+
+public func setAssociatedObject<T>(object: AnyObject, value: T, key: UnsafePointer<Void>, policy: objc_AssociationPolicy) {
```
danielaRiesgo: Why don't we use the same parameter order as the original function?
`public func setAssociatedObject<T>(object: AnyObject, key: UnsafePointer<Void>, value: T, policy: objc_AssociationPolicy)`

Core/Extensions/UIKit/UITextField.swift
```diff
@@ -0,0 +1,26 @@
...
+//  UITextFieldExtension.swift
+//  Core
+//
+//  Created by Francisco Depascuali on 6/29/16.
+//  Copyright © 2016 Wolox. All rights reserved.
+//
+
+import Foundation
+
+private var nextTextFieldKey: UInt8 = 0
```
danielaRiesgo: Move this to the bottom to be consistent.

PR #13, opened 2016-07-25
Adding UIButton extension for underlined title

danielaRiesgo:
## Sumary ##

Made it because from design they said that buttons that navigate to other screens should be underlined like links. So maybe this extension is useful.


woloxci:

FranDepascuali:
👍

PR #14, opened 2016-07-25
String localized receives bundle to use

danielaRiesgo:
## Sumary ##

Making the `localized` String extension receive the bundle to use for localizing.

I found this useful when developing a framework with views, where I didn't want to localize with the main bundle (which is the framework user's one) but the framework's one.

### DIFF
--- a/Core/Extensions/Foundation/String.swift
+++ b/Core/Extensions/Foundation/String.swift
@@ -24,19 +24,10 @@ public extension String {
      Returns a localized representation of the string.
      
      - parameter bundle: Bundle were to search for localization.
-    */
-    public func localized(bundle: NSBundle = NSBundle.mainBundle()) -> String {
-        return NSLocalizedString(self, tableName: nil, bundle: bundle, value: "", comment: "")
-    }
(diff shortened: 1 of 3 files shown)


woloxci:

Core/Extensions/Foundation/String.swift
```diff
@@ -23,17 +23,17 @@ public extension String {
...
     }
     
     /**
      Returns a localized representation of the string.
      
      - parameter arguments: Formatting arguments.
      */
-    public func localized(arguments: CVarArgType...) -> String {
-        return String(format: NSLocalizedString(self, tableName: nil, bundle: NSBundle.mainBundle(), value: "", comment: ""), arguments: arguments)
+    public func localized(bundle: NSBundle = NSBundle.mainBundle(), arguments: CVarArgType...) -> String {
```
FranDepascuali: You could use the other localized

FranDepascuali:
@danielaRiesgo check travis errors also

FranDepascuali:
👍

PR #27, opened 2016-08-11
Adding usage to README

FranDepascuali:
## Summary ##
**Trello card**: https://trello.com/c/lPLPCLpm/14-agregar-usage-al-readme

Adds framework's usage.


woloxci:

README.md
```diff
@@ -38,6 +41,44 @@ github "Wolox/wolmo-core-ios" ~> 0.1.0
 ### Manually
 [Bootstrap](#bootstrap) the project and then drag it to your workspace.
 
+## Usage
+
+The framework is divided into 2 sections: [Utilities](#utilities) and [Extensions](#extensions). 
+
+The main difference between them is that `utilities` represent components that are independent of the existance of another component. On the other hand, `extensions` represent extra functionality for other existent components. Extensions exist only because the components they relies on exists.
```
danielaRiesgo: `they rely on`

README.md
```diff
@@ -38,6 +41,44 @@ github "Wolox/wolmo-core-ios" ~> 0.1.0
...
+4. [AssociatedObject](Core/Utilities/AssociatedObject.swift): Setter and getted for associated objects. Works with objects and value types.
+5. [DeviceType](Core/Utilities/DeviceType.swift): Provides the current device type.
+6. [CellIdentifiable](Core/Utilities/CellIdentifiable.swift): Identifies a cell. Usually used with `UICollectionView` and `UITableView`.
+
+### Extensions
+Extends the behaviour of a component (`class`, `struct`, `enum`, ...) by adding common functionality (functions, computed properties, initializers, ...). 
+
+Extensions contain components which depends exclusively on other components. 
+
+It would not make sense to have these components in [Utilities](#utilities), because they don't exist by themselves, but by adding a behaviour to another existing component.
```
danielaRiesgo: I wouldn't say this because the difference between both is repeated so it's clear. But if you want to put it, I would move it up to the Usage introduction where you talk about both.

README.md
```diff
@@ -38,6 +41,44 @@ github "Wolox/wolmo-core-ios" ~> 0.1.0
...
+
+For example, [Collapsable](Core/Extensions/UIKit/UIView/Collapsable.swift) represents a collapsable element. This behaviour is an addition to a `UIView`; it would not make sense to have something which is not renderable as `Collapsable`. Furthermore, it doesn't make sense for `Collapsable` to exist per se; it is a behaviour added to any view which conforms to it.
```
danielaRiesgo: I wouldn't say this, but if you want to say it at least use another example which is clearer, because I think `Collapsable` is a gray, which could be extension or utility.

README.md
```diff
@@ -38,6 +41,44 @@ github "Wolox/wolmo-core-ios" ~> 0.1.0
...
+
+`Wolmo-core` provides the following extensions:
```
danielaRiesgo: I think the guys said to list all functionalities. So I would list what's inside `ReactiveCocoa`, `AVFoundation`, ... with what they are about in a small description.

README.md
FranDepascuali: I don't think it is a component by itself, but an extension of a UIView. Why do you think it is a gray?

README.md
FranDepascuali: I don't, I think that writing every class would be an overkill, error prone (every time we add a new one we would need to update the README) and also I don't think they are all so important. Also, as the functions provided all work for a class, but they do different things, it would be difficult to describe a functionality as a whole.

README.md
danielaRiesgo: Because anything can be collapsable if you find a meaning to it. Maybe asking a controller to collapse is collapsing its views, or it could be an animation, or if it has child controllers maybe collapsing is using new child controllers. Or maybe a table controller collapses by cleaning its information and being empty. I don't know, examples.
But maybe in the way we described the protocol is not a gray.

README.md
```diff
@@ -38,6 +41,44 @@ github "Wolox/wolmo-core-ios" ~> 0.1.0
...
+
+1. [ReactiveCocoa](Core/Extensions/ReactiveCocoa): Extensions for Signal and SignalProducer.
```
danielaRiesgo: Use "`" with Signal and SignalProducer.

README.md
danielaRiesgo: Not every extension which have its own documentation. But a list of the components for which there is an extension. Not just examples `such as String, Array and NSDate`, but a complete list:
```
Extensions for the following Foundation elements:
`Array`
`NSBundle`
...

```

README.md
FranDepascuali: I don't know either. @guidomb what do you think?

README.md
FranDepascuali: I changed them to links

README.md
FranDepascuali: They are a lot of files. I don't know if they should all be mentioned. What do you think @guidomb ?

danielaRiesgo: At least that was what I understood from the release meeting. Maybe we should ask @epintos

epintos: @danielaRiesgo I think that the more documentation the better. But also we need to take into account that it doesn't deprecate in two seconds. So taking into account that, if we start having lots and lots of classes, maybe its easier to make a little summary of what some folder includes, and that if they want more information, to click in a link.

PR #30, opened 2017-01-03
Swift 3.0 migration

gjcairo:
## Summary ##

**Trello Card:** https://trello.com/c/oyAHTrV3/15-migrar-a-swift-3

This PR migrates Wolmo Core to Swift 3.0


Core/Extensions/Foundation/NSURL.swift
```diff
@@ -28,29 +28,28 @@ public func / (baseURL: NSURL, pathComponent: String) -> NSURL {
...
     /**
      Returns if the URL is of type HTTP protocol.
     */
     public var isHTTPProtocol: Bool {
-        return absoluteString
-                .lowercaseString
-                .substringToIndex(absoluteString.startIndex.advancedBy(5)) // We want HTTP (4) or HTTPS(5)
-                .rangeOfString("http") != nil
+        return absoluteString.lowercased()
+                .substring(to: absoluteString.characters.index(absoluteString.startIndex, offsetBy: 5)) // We want HTTP (4) or HTTPS(5)
```
danielaRiesgo: Apart from the migration, I think the offset could be 4, because even if it is HTTPS, the first 4 would be HTTP and so the range won't be nil.

Core/Extensions/Foundation/NSURL.swift
```diff
@@ -28,29 +28,28 @@ public func / (baseURL: NSURL, pathComponent: String) -> NSURL {
...
+                .range(of: "http") != nil
```
danielaRiesgo: Could we use `.none` here?

Core/Extensions/Foundation/RawRepresentable.swift
```diff
@@ -40,16 +40,16 @@ public extension RawRepresentable where RawValue == Int {
 /**
  Creates a generator for the specified RawValue type.
  */
-public struct RawRepresentableGenerator<RawValue, Representable: RawRepresentable where Representable.RawValue == RawValue>: GeneratorType {
+public struct RawRepresentableGenerator<RawValue, Representable: RawRepresentable>: IteratorProtocol where Representable.RawValue == RawValue {
     
-    private let _valueSuccessor: Representable -> Representable?
-    private var _nextValue: Representable?
+    fileprivate let _valueSuccessor: (Representable) -> Representable?
```
danielaRiesgo: I think these can be `private` directly.

Core/Extensions/Foundation/String.swift
```diff
@@ -28,7 +28,7 @@ public extension String {
      - parameter bundle: Bundle were to search for localization.
      - parameter arguments: Formatting arguments.
      */
-    public func localized(bundle: NSBundle = NSBundle.mainBundle(), arguments: CVarArgType...) -> String {
+    public func localized(_ bundle: Bundle = Bundle.main, arguments: CVarArg...) -> String {
```
danielaRiesgo: Maybe `localized(withBundle: Bundle = Bundle.main, arguments: CVarArg...)` ?

Core/Extensions/Foundation/String.swift
```diff
@@ -28,7 +28,7 @@ public extension String {
...
         return String(format: NSLocalizedString(self, tableName: nil, bundle: bundle, value: "", comment: ""), arguments: arguments)
```
danielaRiesgo: By the way, I don't remember if we have this, but if not:
can we add a card or something to add a `func format(with arguments: CVarArg...) -> String` as an String extension? So we avoid using String(format: "", arguments: ...), it's friendlier.

Core/Extensions/Foundation/String.swift
```diff
@@ -47,16 +47,16 @@ public extension String {
...
     
     /**
      Checks if the string has spaces or not.
      */
     public var hasSpaces: Bool {
-        let whitespace = NSCharacterSet.whitespaceCharacterSet()
-        let range = rangeOfCharacterFromSet(whitespace)
+        let whitespace = CharacterSet.whitespaces
+        let range = rangeOfCharacter(from: whitespace)
         return range != nil
```
danielaRiesgo: Could we use `.none` here?

Core/Extensions/Foundation/String.swift
```diff
@@ -67,14 +67,14 @@ public extension String {
...
-        return emailTest.evaluateWithObject(self)
+        return emailTest.evaluate(with: self)
     }
     
     /*
      Removes leading and trailing whitespaces.
      */
     public var trimmed: String {
-        return stringByTrimmingCharactersInSet(NSCharacterSet.whitespaceCharacterSet())
+        return trimmingCharacters(in: CharacterSet.whitespaces)
```
danielaRiesgo: We could maybe add a card if you think so, to evaluate if we could add the `\n` characters in this `trimmed` or add another function, or do nothing if not.

Core/Extensions/UIKit/UIButton.swift
```diff
@@ -22,13 +22,13 @@ public extension UIButton {
         By default, Normal.
      
      */
-    public func setUnderlinedTitle(title: String, style: NSUnderlineStyle = .StyleSingle, color: UIColor? = .None, forState state: UIControlState = .Normal) {
-        var attributes: [String : AnyObject] = [NSUnderlineStyleAttributeName: style.rawValue]
+    public func setUnderlinedTitle(_ title: String, style: NSUnderlineStyle = .styleSingle, color: UIColor? = .none, forState state: UIControlState = UIControlState()) {
```
danielaRiesgo: Maybe `setUnderlined(title: String` ?

danielaRiesgo: I would prefer to use `UIControlState.normal`

Core/Extensions/UIKit/UITableView.swift
```diff
@@ -15,8 +15,8 @@ public extension UITableView {
      
      - parameter cellType: An identifiable cell to take the identifier from.
      */
-    func registerCell(cellType: IdentifiableCell.Type) {
-        registerNib(UINib(nibName: cellType.cellIdentifier, bundle: nil), forCellReuseIdentifier: cellType.cellIdentifier)
+    func registerCell(_ cellType: IdentifiableCell.Type) {
```
danielaRiesgo: Maybe `func register(cell cellType: IdentifiableCell.Type)` ?

danielaRiesgo: Apart from migration, we could modify this function to receive a `Bundle?` which by default is `.none`.

Core/Extensions/UIKit/UITableView.swift
```diff
@@ -15,8 +15,8 @@ public extension UITableView {
...
+        register(UINib(nibName: cellType.cellIdentifier, bundle: nil), forCellReuseIdentifier: cellType.cellIdentifier)
```
danielaRiesgo: Could we use `.none` here?

Core/Extensions/UIKit/UIView/Collapsable.swift
```diff
@@ -19,15 +19,15 @@ public protocol Collapsable {
      - Parameter animated: Indicates if the collapse should be animated.
      - Parameter animationDuration: The animationDuration of the collapse.
      */
-    func collapse(animated: Bool, animationDuration: NSTimeInterval)
+    func collapse(_ animated: Bool, animationDuration: TimeInterval)
```
danielaRiesgo: I wouldn't add the `_`.

Core/Extensions/UIKit/UIView/Collapsable.swift
```diff
@@ -19,15 +19,15 @@ public protocol Collapsable {
...
     
     /**
      Uncollapse self.
      
      - Parameter animated: Indicates if the collapse should be animated.
      - Parameter animationDuration: The animationDuration of the collapse.
      */
-    func uncollapse(animated: Bool, animationDuration: NSTimeInterval)
+    func uncollapse(_ animated: Bool, animationDuration: TimeInterval)
```
danielaRiesgo: I wouldn't add the `_`.

Core/Extensions/UIKit/UIView/Collapsable.swift
```diff
@@ -45,15 +45,15 @@ extension Collapsable where Self: UIView {
      - Parameter animated: Indicates if the collapse should be animated.
      - Parameter animationDuration: The animationDuration of the collapse.
      */
-    public func collapse(animated: Bool = true, animationDuration: NSTimeInterval = 1) {
+    public func collapse(_ animated: Bool = true, animationDuration: TimeInterval = 1) {
```
danielaRiesgo: I wouldn't add the `_`.

Core/Extensions/UIKit/UIView/Collapsable.swift
```diff
@@ -68,27 +68,27 @@ extension Collapsable where Self: UIView {
      - Parameter animated: Indicates if the collapse should be animated.
      - Parameter animationDuration: The animationDuration of the collapse.
      */
-    public func uncollapse(animated: Bool = true, animationDuration: NSTimeInterval = 1) {
+    public func uncollapse(_ animated: Bool = true, animationDuration: TimeInterval = 1) {
```
danielaRiesgo: I wouldn't add the `_`.

Core/Extensions/UIKit/UIView/Collapsable.swift
```diff
@@ -68,27 +68,27 @@ extension Collapsable where Self: UIView {
...
-                heightConstraint.active = false
+                heightConstraint.isActive = false
             }
         }
         
         updateView(animated, animationDuration: animationDuration)
     }
     
-    private func updateView(animated: Bool, animationDuration: NSTimeInterval) {
+    fileprivate func updateView(_ animated: Bool, animationDuration: TimeInterval) {
```
danielaRiesgo: I think this can be `private`.

Core/Extensions/UIKit/UIView/Collapsable.swift
```diff
@@ -68,27 +68,27 @@ extension Collapsable where Self: UIView {
...
         setNeedsLayout()
         
         if animated {
-            UIView.animateWithDuration(animationDuration) {
+            UIView.animate(withDuration: animationDuration, animations: {
```
danielaRiesgo: A question, all the `method(arg1) { closure }` that were changed to `method(arg1, nameOfArg2: { closure })`, were on purpose? why? I don't know if you don't like that way, I got used to it and prefer it now because it uses less parenthesis that you don't know where they come from.

Core/Extensions/UIKit/UIView/Collapsable.swift
```diff
@@ -99,8 +99,8 @@ private var CollapsableHeightConstraintKey: UInt8 = 0
 
 private extension UIView {
     
-    private var previousHeightConstraint: NSLayoutConstraint? {
-        return constraints.filterFirst { $0.firstAttribute == .Height }
+    var previousHeightConstraint: NSLayoutConstraint? {
```
danielaRiesgo: I wouldn't eliminate the `private` here. Although I think that this extension should be `fileprivate`.

Core/Extensions/UIKit/UIView/NibLoadable.swift
```diff
@@ -25,8 +25,8 @@ extension NibLoadable where Self: UIView {
         - parameter bundle: Specific bundle, default = bundle for the class.
         - returns: The loaded UIView
     */
-    public static func loadFromNib(bundle: NSBundle = NSBundle(forClass: Self.self)) -> Self? {
-        let nibName = NSStringFromClass(self).componentsSeparatedByString(".").last
+    public static func loadFromNib(_ bundle: Bundle = Bundle(for: Self.self)) -> Self? {
```
danielaRiesgo: `func loadFromNib(inBundle bundle: Bundle = Bundle(for: Self.self)) -> Self?` ?

Core/Extensions/UIKit/UIView/UIView.swift
```diff
@@ -95,19 +95,19 @@ extension UIView {
      - parameter containerView: The container view.
      - parameter viewPositioning: Back or Front. Default: Front
      */
-    public func loadInto(containerView: UIView, viewPositioning: ViewPositioning = .Front) {
+    public func loadInto(_ containerView: UIView, viewPositioning: ViewPositioning = .front) {
```
danielaRiesgo: Maybe `load(into containerView: UIView, ...)` ?

Core/Extensions/UIKit/UIView/UIView.swift
```diff
@@ -34,7 +34,7 @@ extension UIView {
      
      Note: By default, the border will have the width of the view.
      */
-    public func topBorder(border: Border, offset: CGFloat = 0) {
+    public func topBorder(_ border: Border, offset: CGFloat = 0) {
```
danielaRiesgo: Maybe `top(border: Border, ...)` ?

Core/Extensions/UIKit/UIView/UIView.swift
```diff
@@ -49,7 +49,7 @@ extension UIView {
      
      Note: By default, the border will have the width of the view.
      */
-    public func bottomBorder(border: Border, offset: CGFloat = 0) {
+    public func bottomBorder(_ border: Border, offset: CGFloat = 0) {
```
danielaRiesgo: Maybe `bottom(border: Border, ...)` ?

Core/Extensions/UIKit/UIView/UIView.swift
```diff
@@ -65,7 +65,7 @@ extension UIView {
      
      Note: By default, the border will have the height of the view.
      */
-    public func leftBorder(border: Border, offset: CGFloat = 0) {
+    public func leftBorder(_ border: Border, offset: CGFloat = 0) {
```
danielaRiesgo: Maybe `left(border: Border, ...)` ?

Core/Extensions/UIKit/UIView/UIView.swift
```diff
@@ -80,7 +80,7 @@ extension UIView {
      
      Note: By default, the border will have the height of the view.
      */
-    public func rightBorder(border: Border, offset: CGFloat = 0) {
+    public func rightBorder(_ border: Border, offset: CGFloat = 0) {
```
danielaRiesgo: Maybe `right(border: Border, ...)` ?

Core/Extensions/UIKit/UIViewController.swift
```diff
@@ -34,10 +34,10 @@ public extension UIViewController {
      - parameter into: The containerView into which the controller will be loaded.
      - parameter viewPositioning: Back or Front. Default: Front
      */
-    public func loadViewController(childViewController: UIViewController, into containerView: UIView, viewPositioning: ViewPositioning = .Front) {
-        childViewController.willMoveToParentViewController(self)
+    public func loadViewController(_ childViewController: UIViewController, into containerView: UIView, viewPositioning: ViewPositioning = .front) {
```
danielaRiesgo: Maybe `load(viewcontroller childViewController: UIViewController, ...)` ?

Core/Extensions/UIKit/UIViewController.swift
```diff
@@ -52,9 +52,9 @@ public extension UIViewController {
 
 private extension UIViewController {
     
-    private func setHiddenValueFromBarsTo(value: Bool) {
-        tabBarController?.tabBar.hidden = value
-        UIApplication.sharedApplication().statusBarHidden = value
+    func setHiddenValueFromBarsTo(_ value: Bool) {
```
danielaRiesgo: I can write this in the correct line, but I think the get from the `hideBars` property is not correct. Maybe add a card to check it.

danielaRiesgo: Maybe `setHiddenValueFromBars(to value: Bool)` or directly `setBarHidden(to value: Bool)` ?

Core/Extensions/UIKit/UIViewController.swift
```diff
@@ -52,9 +52,9 @@ public extension UIViewController {
...
+        tabBarController?.tabBar.isHidden = value
+        UIApplication.shared.isStatusBarHidden = value
```
danielaRiesgo: Couldn't this also use `navigationController?.navigationBar.isHidden` to be consistent?

Core/Utilities/ActionHandler.swift
```diff
@@ -11,12 +11,17 @@
  */
 public protocol ActionHandlerType {
     
-    func setAction(events: UIControlEvents, _ action: (Self, UIEvent) -> Void)
+    associatedtype SenderType
```
danielaRiesgo: Why did you change this? I prefer the other way around, it avoids associated types.

Core/Utilities/ActionHandler.swift
```diff
@@ -25,26 +30,24 @@ extension ActionHandlerType where Self: UIControl {
      
      - note: It retains any object used in the closure.
      */
-    public func setAction(events: UIControlEvents = .TouchUpInside, _ action: (Self, UIEvent) -> Void) {
+    public func setAction(_ events: UIControlEvents = .touchUpInside, _ action: @escaping (UIControl, UIEvent) -> Void) {
```
danielaRiesgo: Maybe `func setAction(for events: ...)` ?
Also I prefer leaving it in `Self` instead of `UIControl`.

Core/Utilities/ActionHandler.swift
```diff
@@ -25,26 +30,24 @@ extension ActionHandlerType where Self: UIControl {
...
 }
 
-extension UIControl: ActionHandlerType {}
-
-private let actionHandlerTypeAssociatedObjectKey = UnsafeMutablePointer<Int8>.alloc(1)
+private let actionHandlerTypeAssociatedObjectKey = UnsafeMutablePointer<Int8>.allocate(capacity: 1)
 
 private class ActionHandler<T>: NSObject {
-    private let _action: ((T, UIEvent) -> Void)
+    fileprivate let _action: ((T, UIEvent) -> Void)
```
danielaRiesgo: I think `private` is okay.

Core/Utilities/Alerts/AlertViewPresenter.swift
```diff
@@ -13,10 +13,10 @@ import Foundation
  */
 public protocol AlertViewPresenter {
 
-    func presentAlert(alertViewModel: ErrorAlertViewModel, completion: (() -> Void)?)
+    func presentAlert(_ alertViewModel: ErrorAlertViewModel, completion: (() -> Void)?)
```
danielaRiesgo: Maybe `func presentAlert(viewModel: ...)` ? Although I'm not so sure in this one.

danielaRiesgo: Or maybe `present(alert alertViewModel: ...)` similar to what they have now.

Core/Utilities/AssociatedObject.swift
```diff
@@ -18,17 +18,12 @@
  - warning: Using associated objects should be avoided, as they are used in runtime and not in compile time, and could lead to performance issues.
  - seealso: objc_setAssociatedObject.
  */
-public func setAssociatedObject<T>(object: AnyObject, key: UnsafePointer<Void>, value: T?, policy: objc_AssociationPolicy = .OBJC_ASSOCIATION_RETAIN) {
+public func setAssociatedObject<T>(_ object: AnyObject, key: UnsafeRawPointer, value: T?, policy: objc_AssociationPolicy = .OBJC_ASSOCIATION_RETAIN) {
```
danielaRiesgo: Maybe `set<T>(associatedObject: AnyObject, ...)` ?

Core/Utilities/AssociatedObject.swift
```diff
@@ -41,13 +36,13 @@ public func setAssociatedObject<T>(object: AnyObject, key: UnsafePointer<Void>,
  - warning: Using associated objects should be avoided, as they are used in runtime and not in compile time, and could lead to performance issues.
  - seealso: objc_getAssociatedObject.
  */
-public func getAssociatedObject<T>(object: AnyObject, key: UnsafePointer<Void>) -> T? {
+public func getAssociatedObject<T>(_ object: AnyObject, key: UnsafeRawPointer) -> T? {
```
danielaRiesgo: Same here

Core/Utilities/Alerts/AlertViewPresenter.swift
```diff
@@ -31,9 +31,9 @@ extension UIViewController: AlertViewPresenter {
      Dismiss and confirmation callbacks can be set on the `ErrorAlertViewModel`.
      - Seealso: `ErrorAlertViewModel`
      */
-    public func presentAlert(alertViewModel: ErrorAlertViewModel, completion: (() -> Void)? = .None) {
+    public func presentAlert(_ alertViewModel: ErrorAlertViewModel, completion: (() -> Void)? = .none) {
```
danielaRiesgo: Same here.

Core/Utilities/Alerts/AlertViewPresenter.swift
```diff
@@ -13,10 +13,10 @@ import Foundation
  */
 public protocol AlertViewPresenter {
 
-    func presentAlert(alertViewModel: ErrorAlertViewModel, completion: (() -> Void)?)
+    func presentAlert(_ alertViewModel: ErrorAlertViewModel, completion: (() -> Void)?)
     
 
-    func presentAlert(alertViewModel: ConfirmationAlertViewModel, completion: (() -> Void)?)
+    func presentAlert(_ alertViewModel: ConfirmationAlertViewModel, completion: (() -> Void)?)
```
danielaRiesgo: Same here.

Core/Utilities/AssociatedObject.swift
```diff
@@ -69,6 +64,6 @@ private final class AssociatedObjectBox<T> {
  
  - returns: An AssociatedObjectBox (class) which holds the value.
  */
-private func lift<T>(value: T) -> AssociatedObjectBox<T> {
+private func lift<T>(_ value: T) -> AssociatedObjectBox<T> {
```
danielaRiesgo: I think this function isn't needed anymore, and so the `AssociatedObjectBox`.

Core/Utilities/CellIdentifiable.swift
```diff
@@ -21,6 +21,6 @@ extension UITableViewCell: IdentifiableCell {
     /**
      Returns the identifier of the class. It is the name of the class.
      */
-    public static var cellIdentifier: String { return String(self) }
+    public static var cellIdentifier: String { return String(describing: self) }
```
danielaRiesgo: Let's add this extension for all `UICollectionViewCell`s instead of only UITableViewCells.

danielaRiesgo: Moreover, I think this does not work as it was before, so I would check this over. We want to have this:
if the cell class is called `SomeCell`, `cellIdentifier` should be `SomeCell`.
But for what I saw in the documentation of that init, this is not the case, but it describes the class name and attributes.

Core/Utilities/ProgressHUD/ProgressHUDDelegate.swift
```diff
@@ -19,7 +19,7 @@ public protocol ProgressHUDDelegate {
      
      - parameter labelText: A label text to be shown under the progress hud.
     */
-    func showProgressHud(labelText: String?)
+    func showProgressHud(_ labelText: String?)
```
danielaRiesgo: `func showProgressHud(withText text: String?)` ?

Core/Utilities/ProgressHUD/ProgressHUDDelegate.swift
```diff
@@ -35,19 +35,19 @@ extension UIViewController: ProgressHUDDelegate {
      - parameter labelText: A label text to be shown under the progress hud.
      - Warning: Showing a progress hud freezes the viewController's view.
      */
-    public func showProgressHud(labelText: String? = .None) {
-        let progressHUD = MBProgressHUD.showHUDAddedTo(view, animated: true)
+    public func showProgressHud(_ labelText: String? = .none) {
```
danielaRiesgo: Same here.

Core/Utilities/ProgressHUD/ProgressHUDDelegate.swift
```diff
@@ -35,19 +35,19 @@ extension UIViewController: ProgressHUDDelegate {
...
+        let progressHUD = MBProgressHUD.showAdded(to: view, animated: true)
         
         if let labelText = labelText {
-            progressHUD.labelText = labelText
+            progressHUD?.labelText = labelText
```
danielaRiesgo: We could add a card to evaluate why this is an optional and what should the method do if you expect a hud to be shown and you can't, should it do nothing? should it explode?

CoreTests/Extensions/Foundation/ArraySpec.swift
```diff
@@ -12,9 +12,9 @@ import Quick
 import Nimble
 import Core
 
-public class ArraySpec: QuickSpec {
+open class ArraySpec: QuickSpec {
```
danielaRiesgo: Don't know why tests should be `open` instead of `public`. Your call.

Core/Extensions/Foundation/NSBundle.swift
```diff
@@ -17,8 +17,8 @@ public extension NSBundle {
      - returns: The loaded NibType or .None if it don't exists.
      - seealso: loadNibNamed()
      */
-    public func loadNib<NibType, T: RawRepresentable where T.RawValue == String>(nibName: T) -> NibType? {
-        return loadNibNamed(nibName.rawValue, owner: self, options: nil)[0] as? NibType
+    public func loadNib<NibType, T: RawRepresentable>(_ nibName: T) -> NibType? where T.RawValue == String {
```
FranDepascuali: I think that it should be ```loadNibNamed()``` as it isn't a preposition (and also it respects the original function name which is being called `loadNibNamed`)

Cartfile
```diff
@@ -1,2 +1,2 @@
-github "ReactiveCocoa/ReactiveCocoa" ~> 4.1.0
+github "ReactiveCocoa/ReactiveCocoa" "5.0.0-alpha.5"
 github "jdg/MBProgressHUD" ~> 0.9.2
```
gjcairo: This isn't the latest version (1.0), so I could update it anyway, doesn't hurt. This seems to work fine though.

gjcairo: Took a closer look and the only change in version 1.0.0 seems to be an updated README file 🤔  I can still update it.

Core/Extensions/Foundation/Array.swift
```diff
@@ -17,7 +17,7 @@ public extension Array {
      
      - returns: Dictionary with elements grouped by their corresponding keys.
      */
-    public func groupBy<K>(buildGroupKey: Element -> K) -> [K: [Element]] {
+    public func groupBy<K>(_ buildGroupKey: (Element) -> K) -> [K: [Element]] {
```
gjcairo: Hmm, what about `groupElements<K>(usingCriteria keyForElement: (Element) -> K)`?

danielaRiesgo: `byCriteria` or `withCriteria` to be shorter? Any of them is okay for me.

Core/Extensions/Foundation/Array.swift
```diff
@@ -69,8 +69,8 @@ public extension Array {
      - returns: The element that satisfies the predicate or .None if none satisfies.
      - seealso: indexOf()
     */
-    public func filterFirst(condition: Element -> Bool) -> Element? {
-        return indexOf(condition).map { self[$0] }
+    public func filterFirst(_ condition: (Element) -> Bool) -> Element? {
```
gjcairo: Given that what this function does is actually return the first element that satisfies the given condition, I think a better name for it would be `getFirst(where condition: (Element) -> Bool)`. What do you think? I'll also update the docs.

Core/Extensions/Foundation/Date.swift
```diff
@@ -73,8 +73,8 @@ public extension NSDate {
      retrieve both components at the same time.
      - returns: The requested date components.
      */
-    public func components(units: NSCalendarUnit, using calendar: NSCalendar = .currentCalendar()) -> NSDateComponents {
-        return calendar.components(units, fromDate: self)
+    public func components(_ units: Set<Calendar.Component>, using calendar: Calendar = .current) -> DateComponents {
```
gjcairo: I did change the docs, but I don't think we should update the name of the function for the reason you mentioned: it's a `Date` extension, so you would use it like `date.components(...)`, and I think this falls in line better with the new syntax than `date.dateComponents(...)` (i.e., `String().localizedString` has been changed to `String().localized`).

Core/Extensions/Foundation/NSTimer.swift
```diff
@@ -20,11 +20,11 @@ public extension NSTimer {
      - seealso: CFRunLoopTimerCreateWithHandler()
      - returns: The newly-created `NSTimer` instance.
      */
-    public static func schedule(delay: NSTimeInterval, handler: NSTimer! -> Void) -> NSTimer {
+    public static func schedule(_ delay: TimeInterval, handler: @escaping (CFRunLoopTimer?) -> Void) -> Timer {
```
gjcairo: I think the best approach (from what I understand of the new naming conventions) would be `schedule(withDelay delay: TimeInterval, handler: @escaping (CFRunLoopTimer?) -> Void)`

Core/Extensions/Foundation/NSTimer.swift
```diff
@@ -20,11 +20,11 @@ public extension NSTimer {
...
         let fireDate = delay + CFAbsoluteTimeGetCurrent()
         let timer = CFRunLoopTimerCreateWithHandler(kCFAllocatorDefault, fireDate, 0, 0, 0, handler)
-        CFRunLoopAddTimer(CFRunLoopGetCurrent(), timer, kCFRunLoopCommonModes)
-        return timer
+        CFRunLoopAddTimer(CFRunLoopGetCurrent(), timer, CFRunLoopMode.commonModes)
+        return timer!
```
gjcairo: `CFRunLoopTimerCreateWithHandler()` returns a `CFRunLoopTimer!`: an _[implicitly unwrapped optional](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/TheBasics.html#//apple_ref/doc/uid/TP40014097-CH5-ID334)_. When assigning its return value to `let timer`, the inferred type of `timer` is (correctly) `Timer?`, because it's an optional but ignores the "implicitly unwrapped" property. `CFRunLoopTimerCreateWithHandler()` guarantees that its return value isn't `nil`, so we could either return `timer!` like we're doing here, or we could change the assignment to `timer` to:
```swift
let timer: Timer = CFRunLoopTimerCreateWithHandler(kCFAllocatorDefault, fireDate, 0, 0, 0, handler)
```
This works because the return value is implicitly unwrapped.

Core/Extensions/Foundation/NSTimer.swift
```diff
@@ -39,11 +39,11 @@ public extension NSTimer {
...
      - returns: The newly-created `NSTimer` instance.
      */
-    public static func schedule(repeatInterval interval: NSTimeInterval, handler: NSTimer! -> Void) -> NSTimer {
+    public static func schedule(repeatInterval interval: TimeInterval, handler: @escaping (CFRunLoopTimer?) -> Void) -> Timer {
         let fireDate = interval + CFAbsoluteTimeGetCurrent()
         let timer = CFRunLoopTimerCreateWithHandler(kCFAllocatorDefault, fireDate, interval, 0, 0, handler)
-        CFRunLoopAddTimer(CFRunLoopGetCurrent(), timer, kCFRunLoopCommonModes)
-        return timer
+        CFRunLoopAddTimer(CFRunLoopGetCurrent(), timer, CFRunLoopMode.commonModes)
+        return timer!
```
gjcairo: See above.

Core/Extensions/Foundation/NSTimer.swift
danielaRiesgo: I like the second one better.

Core/Extensions/Foundation/NSURL.swift
gjcairo: Yes, you're right. However, I think maybe a little refactor is needed here: the string passed is being `lowercased()` and also a new string is being created with `substring(to:)`. All this is unnecessary, since the only thing that we're using these strings for is for a simple search - and I'm not only talking about time complexity here, but mostly about the ease with which this function is read. 

Maybe something like this is a little easier on the eyes (and shorter / faster, complexity-wise):
```swift
public var isHTTPProtocol: Bool {
    let httpRange = absoluteString.startIndex ..< absoluteString.characters.index(absoluteString.startIndex, offsetBy: 4)
    return absoluteString.range(of: "http", options: .caseInsensitive, range: httpRange) != nil
    }
```

Core/Extensions/Foundation/String.swift
gjcairo: Following [Swift's guidelines](https://swift.org/documentation/api-design-guidelines/#naming), I think the best approach would be something like:

```swift
public func localized(withArguments arguments: CVarArg..., bundle: Bundle = Bundle.main) -> String
```
- Optional arguments should be listed at the end of the parameters list, so the `bundle` argument should be swapped with the `arguments` arg.
- As you mentioned, function names should read like phrases, and that includes adding labels with the corresponding prepositions before arguments that should be read as part of said phrases.

guidomb: Are you sure? I would have written

```swift
public func localized(with arguments: CVarArg..., bundle: Bundle = Bundle.main) -> String
```

guidomb: That being said, in this particular case that wouldn't be possible because (I think) variable arguments should be the last parameter.

Core/Extensions/Foundation/String.swift
gjcairo: Done

Core/Extensions/Foundation/String.swift
gjcairo: I wouldn't change `trimmed` to not-ignore newlines: they're whitespace and, like the docs say, this removes leading and trailing whitespace - that includes newlines. I don't know what use a different `trimmedWithNewlines` property would have, but if you've come across this use case before, we could create it.

gjcairo: By the way, I think we should change the "Removes leading and trailing whitespace" to something more like "Returns a copy of the string without its leading and trailing whitespace", as this makes it clearer that you're not mutating the string. Thoughts?

Core/Extensions/UIKit/UITableView.swift
gjcairo: Hmm, `register(cellType:)`?

gjcairo: You mean something like `func register(cellType: IdentifiableCell.Type, bundle: Bundle? = .none) `?

Core/Extensions/UIKit/UITableView.swift
gjcairo: If we passed the bundle as an argument then there's no need for this.

Core/Extensions/UIKit/UIView/Collapsable.swift
gjcairo: This was changed automatically by Xcode when transitioning to Swift 3 syntax. I can change it back to what it was.

Core/Extensions/UIKit/UIView/Collapsable.swift
gjcairo: Added `fileprivate`. I also had to add the modifier to the extension itself (changed it from `private`).

Core/Extensions/UIKit/UIView/UIView.swift
gjcairo: Hmm, I'm not too sure about naming for this... maybe addTopBorder(border:offset:)?

Also, I think this could be more generic / its result is kind of confusing. As it is, the border will always begin from the top left corner of the view, and the "offset" is actually the width. I think this would be better if the offset was _really_ an offset, that is, if you could specify from which value of X you want your border to appear. You could pass a positive value if you want the border to start to the left of the origin of the view, and a negative one if you want it to end before the width of the view (as it's working now).

It would look something like this:
```swift
public func addTopBorder(border: Border, offset: CGFloat = 0) {
    let offsetX = max(0, offset)
    let borderView = UIView(frame: CGRect(x: offsetX, y: 0, width: bounds.width - fabs(offset), height: CGFloat(border.height)))
    borderView.backgroundColor = border.color
    addSubview(borderView)
}
```

What do you think about this? I think this could also apply to the other three methods. I can create a card to refactor this so as not to miss the point of this PR, but I think we should do this.

Core/Extensions/UIKit/UIView/UIView.swift
gjcairo: See `topBorder()`

Core/Extensions/UIKit/UIView/UIView.swift
gjcairo: See `topBorder()`

Core/Extensions/UIKit/UIView/UIView.swift
gjcairo: See `topBorder()`

Core/Extensions/UIKit/UIView/UIView.swift
gjcairo: I agree with using `into` as the label, but don't you think `add(into containerView: UIView, ...)` is better? I mean, you `addSubview()`s, you don't load them. It sounds weird to me for some reason.

Core/Extensions/UIKit/UIViewController.swift
gjcairo: The name of the function + the argument form a grammatical phrase, not a prepositional one. Thus, following [naming conventions](https://swift.org/documentation/api-design-guidelines/#naming), we should omit the argument's label.

I'd name it `loadChildViewController(_ viewController: UIViewController, ...)` because it's more obvious that you're adding a child view controller to the current view controller.

Core/Extensions/UIKit/UIViewController.swift
gjcairo: Hmm yeah, looks weird. I added the card.

gjcairo: What about `func setBarsHidden(_ hidden: Bool)`?

Core/Extensions/UIKit/UIViewController.swift
gjcairo: `statusBar` != `navigationBar`: the status bar is where the reception/carrier/clock/battery info appears.

Core/Utilities/ActionHandler.swift
gjcairo: This had to be changed because the use of `Self` is not permitted for non-final classes anymore. UIControl is not a final class, so in its extension we got a compiler error.

Core/Extensions/Foundation/NSURL.swift
danielaRiesgo: Great ! But use `.none` if it can be used instead of `nil`.

Core/Utilities/ActionHandler.swift
gjcairo: What about `setAction(forEvents events: UIControlEvents = .touchUpInside, ...)`?

Also, see previous answer about `Self`.

Core/Extensions/Foundation/String.swift
danielaRiesgo: I guess the variable arguments may be at the beginning if we put an obligatory label for the bundle parameter, because then everything before it would be argument.
And I guess `localized(with arguments: CVarArg...` is simpler, although `localized(withArguments arguments: CVarArg...` may be friendlier for someone that doesn't know the function.

Core/Utilities/Alerts/AlertViewPresenter.swift
gjcairo: `presentAlert(with viewModel: ErrorAlertViewModel, ...)`?

Core/Utilities/Alerts/AlertViewPresenter.swift
gjcairo: Same as above.

Core/Utilities/Alerts/AlertViewPresenter.swift
gjcairo: Same as above.

Core/Extensions/Foundation/String.swift
guidomb: Are you sure? Not really sure if that compiles. Either case I've never seen a method with variable arguments as the first argument.

Core/Utilities/AssociatedObject.swift
gjcairo: The argument doesn't forma prepositional phrase with the function name, so I think that the name it has right now is fine.

Core/Utilities/AssociatedObject.swift
gjcairo: Same as above.

Core/Extensions/Foundation/String.swift
danielaRiesgo: First comment: Then I guess we should use here `CharacterSet.whitespacesAndNewlines`.
Second comment: I think there is a convention like: if the verb ends in `ing` it returns a new string, if it is in past tense then it modifies the current string. Like `append` and `appending`. Maybe we should call this `trimming` ?

Core/Extensions/UIKit/UITableView.swift
danielaRiesgo: Right

Core/Extensions/UIKit/UIView/UIView.swift
danielaRiesgo: Maybe we should add 2 offsets, left and right (or bottom and top). Because, if I remember correctly, all borders had the offset at a side, that would, if executing all 4 border functions, no have any overlapping border views.
I guess it will be clearer having a left and right border, both optional, and letting the user decide how to assemble border views if he doesn't want them to overlap.
Moreover, I think I wouldn't oblige the offset to be zero or positive, since maybe he wants the border to standout a little bit on both sides. So, I would take the offset as it comes and I would add in the documentation a comment like `If you'd like to use negative offsets, review that this view doesn't have the clipToBounds property on.`
So it would be something like:
```
public func addTopBorder(border: Border, leftOffset: CGFloat = 0, rightOffset: CGFloat = 0) {
    let borderView = UIView(frame: CGRect(x: leftOffset, y: 0, width: bounds.width - leftOffset - rightOffset, height: CGFloat(border.height)))
    borderView.backgroundColor = border.color
    addSubview(borderView)
}
```

Apart from this, if you can find a better name than `height` for the border's property great. Because height would be for horizontal borders, and width for vertical borders.

Core/Utilities/ProgressHUD/ProgressHUDDelegate.swift
gjcairo: The type is optional because `showAdded()` returns an implicitly unwrapped optional - it's the same thing I explained in a previous comment. I think we should resolve it the same way we did there:
```swift
let progressHUD: MBProgressHUD = MBProgressHUD.showAdded(to: view, animated: true)
```

Core/Extensions/UIKit/UIView/UIView.swift
danielaRiesgo: Sure. I think it got caught up with the `loadChildViewController` functions.

CoreTests/Extensions/Foundation/ArraySpec.swift
gjcairo: I think Xcode automatically changed it because what used to be `public` is now `open` (public is now more restrictive). I'll change it to public.

Core/Extensions/UIKit/UIViewController.swift
danielaRiesgo: Sure

Core/Extensions/UIKit/UIViewController.swift
danielaRiesgo: Right ! Sorry

Core/Utilities/Alerts/AlertViewPresenter.swift
danielaRiesgo: Hmm, I don't know. I'm not sure of any of those :P
@guidomb what do you think?

Core/Extensions/Foundation/String.swift
danielaRiesgo: The only bad thing of the bundle first, is that it's true that is optional and like "less important". And if we do want to pass a value to that the naming of the function will be weird, at least for me, or less nice.
`public func localized(bundle: Bundle = Bundle.main, withArguments arguments: CVarArg...) -> String`
`public func localized(in bundle: Bundle = Bundle.main, with arguments: CVarArg...) -> String`
`public func localized(withBundle bundle: Bundle = Bundle.main, arguments: CVarArg...) -> String`
`public func localized(with bundle: Bundle = Bundle.main, arguments: CVarArg...) -> String`

Core/Utilities/CellIdentifiable.swift
gjcairo: It works as intended.

Core/Extensions/Foundation/String.swift
danielaRiesgo: And in the playground it compiles with bundle at the end and works well, only if bundle has a label and you use it when calling it to avoid using the default one.

Core/Extensions/Foundation/Array.swift
gjcairo: I like `withCriteria` better.

Core/Extensions/Foundation/String.swift
gjcairo: Yes, it works fine if you add a label to the subsequent parameters. I think swapping them is better not only because `bundle` is optional and should go at the end of the arguments list, but because the "really important" part of the method should come right after the name of the function - and in this case, it's the arguments list.

To quote [the guidelines](https://swift.org/documentation/api-design-guidelines/#naming):
> Prefer to locate parameters with defaults toward the end of the parameter list. Parameters without defaults are usually more essential to the semantics of a method, and provide a stable initial pattern of use where methods are invoked.

Core/Extensions/Foundation/String.swift
gjcairo: 1. 👍 
2. The convention is -ing or -ed for nonmutating functions/methods, and imperative for their mutating counterparts.

Core/Extensions/UIKit/UIView/UIView.swift
gjcairo: Created the card.

Core/Extensions/Foundation/Array.swift
```diff
@@ -62,14 +62,14 @@ public extension Array {
     }
     
     /**
-     Returns the first index where predicate returns true for the corresponding value, 
-     or nil if such value is not found.
+     Returns the first element of the array for which the predicate returns true,
+     or nil if there is no such element.
```
danielaRiesgo: `or .none`

danielaRiesgo:
Approved this far but don't forget we still haven't decided on how to name `presentAlert`.

Core/Utilities/Alerts/AlertViewPresenter.swift
guidomb: Personally I would go with `present(alert alertViewModel: ...)` but I don't have any strong arguments beside that it reads well for me.

woloxci:

Core/Utilities/Alerts/AlertViewPresenter.swift
gjcairo: Naming conventions don't say the phrase "presentAlert" should be split, because it's not a prepositional phrase, that's why I proposed `presentAlert(with viewModel: ...)` (which adds the preposition so we can have the label). An alternative without adding the "with" would be to leave it as it is, `presentAlert(_ alertViewModel: ...)`. We could make an exception and name it like @guidomb proposes but I don't think the split is necessary there.

danielaRiesgo: Okay. Then I vote for `presentAlert(_ alertViewModel: ...)`

gjcairo:
I'll merge now so I can proceed with `WolmoAuthentication`'s migration to Swift 3. I'll add a Trello card to fix Travis.

woloxci:

PR #33, opened 2017-02-09
Fixing NibLoadable

danielaRiesgo:
## Summary ##

Fixing NibLoadable protocol since in the current Swift, `Self` related functions can't be implemented by non-final classes.
And making extension public.


woloxci:

PR #35, opened 2017-02-17
Adding functions to relate strings and dates

danielaRiesgo:
## Summary ##
Adding date from string and to string functions in NSDate.

## Known issues ##
If we want to give the possibility of being able to specify the format with a date formatter or with a string, and have default values for both, we have a naming problem, because if we do `toString()` it's ambiguous.
I guess we could have a default formatter but not a default format, since the default format is taken from the default formatter,and if you want that, you can use the default formatter function and that's it.


pablog:
I agree with the way of solving the names collapsing. 👍

woloxci:

PR #36, opened 2017-02-22
Adding string functions and Date.now()

danielaRiesgo:
## Summary ##

Adding functions to string and date extension.

These functions where taken from GreenCommuter when seeing they were needed for training.

### DIFF
--- a/Core/Extensions/Foundation/String.swift
+++ b/Core/Extensions/Foundation/String.swift
@@ -104,7 +104,7 @@ public extension String {
     /**
      Returns true if the string is not empty, false if not.
     */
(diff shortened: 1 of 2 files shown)


woloxci:

Core/Extensions/Foundation/String.swift
```diff
@@ -101,4 +101,46 @@ public extension String {
         return replacing(" ", with: "")
     }
     
+    /**
+     Returns true if the string is not empty, false if not.
+    */
+    public func isNotEmpty() -> Bool {
```
pablog: Wouldn't you make it a property too?

PR #37, opened 2017-02-24
Adding dictionary appending functions

danielaRiesgo:
## Summary ##

Adding a function to append dictionary contents (came up in training).
Also associating Extension folders to the file systems folders (that's why the changes in the project file).

### DIFF
--- a/Core/Extensions/Foundation/Dictionary.swift
+++ b/Core/Extensions/Foundation/Dictionary.swift
@@ -15,9 +15,9 @@ extension Dictionary {
         If both dictionaries have the same key with different value, the result
         will have that key with the value found in `dictionary`.
     */
-    func append(contentsOf dictionary: Dictionary<Key, Value>) -> Dictionary<Key, Value> {
+    func appending(contentsOf dictionary: Dictionary<Key, Value>) -> Dictionary<Key, Value> {
         var result = self
-        result.appending(contentsOf: dictionary)
+        result.append(contentsOf: dictionary)
         return result
     }
     
@@ -26,9 +26,9 @@ extension Dictionary {
          If both dictionaries have the same key with different value,
          the result will have that key with the value found in `dictionary`.
      */
-    mutating func appending(contentsOf dictionary: Dictionary<Key, Value>) {
+    mutating func append(contentsOf dictionary: Dictionary<Key, Value>) {
         for (key,value) in dictionary {
-            updateValue(value, forKey:key)
+            updateValue(value, forKey: key)
         }
     }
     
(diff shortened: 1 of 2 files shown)


woloxci:

Core/Extensions/Foundation/Dictionary.swift
```diff
@@ -0,0 +1,35 @@
...
+        result.appending(contentsOf: dictionary)
+        return result
+    }
+    
+    /**
+         Adds to self the contents of the given `dictionary`.
+         If both dictionaries have the same key with different value,
+         the result will have that key with the value found in `dictionary`.
+     */
+    mutating func appending(contentsOf dictionary: Dictionary<Key, Value>) {
```
pablog: Why `append(contentsOf dictionary: Dictionary<Key, Value>)` returns a new dictionary? and this one appends in the current one?

danielaRiesgo: It should be the other way around, right?

Core/Extensions/Foundation/Dictionary.swift
```diff
@@ -0,0 +1,35 @@
...
+extension Dictionary {
+    
+    /**
+        Returns a new dictionary with the contents of self and the given `dictionary`.
+        If both dictionaries have the same key with different value, the result
+        will have that key with the value found in `dictionary`.
+    */
+    func appending(contentsOf dictionary: Dictionary<Key, Value>) -> Dictionary<Key, Value> {
+        var result = self
+        result.append(contentsOf: dictionary)
```
damfinkel: Does swift know that the Dictionary that receives the message must be Dictionary<Key, Value> ?

Core/Extensions/Foundation/Dictionary.swift
```diff
@@ -0,0 +1,35 @@
...
+
+extension Dictionary {
+    
+    /**
+        Returns a new dictionary with the contents of self and the given `dictionary`.
+        If both dictionaries have the same key with different value, the result
+        will have that key with the value found in `dictionary`.
+    */
+    func appending(contentsOf dictionary: Dictionary<Key, Value>) -> Dictionary<Key, Value> {
+        var result = self
```
damfinkel: why is this necessary?

danielaRiesgo: To create a copy of the value type, since the append function is mutating.

Core/Extensions/Foundation/Dictionary.swift
danielaRiesgo: Yes, since here `Key` and `Value` are typealias that exists in the `Dictionary` type (see that the function is not generic) and we can reach them since we are "inside" the `Dictionary` type in the extension.

PR #38, opened 2017-03-03
Adding appending functions for urls

danielaRiesgo:
## Summary ##
Adding "url appending" functions for string and numbers (int and uint).

### DIFF
--- a/Core/Extensions/Foundation/String.swift
+++ b/Core/Extensions/Foundation/String.swift
@@ -8,8 +8,6 @@
 
 import Foundation
 
-
-
 public extension String {
     
     /**
--- a/Core/Extensions/Foundation/String.swift
+++ b/Core/Extensions/Foundation/String.swift
@@ -8,41 +8,7 @@
 
 import Foundation
 
-/**
-    Appends two string, similar to web resource addressing.
-    
-    - parameter basePath: base string.
-    - parameter pathComponent: string that will be appended.
- 
-    - returns: The new NSURL, with pathComponent appended to baseURL.
- */
-public func / (basePath: String, pathComponent: String) -> String {
-    return (basePath as NSString).appendingPathComponent(pathComponent)
-}
 
-/**
-     Appends an integer to a string, similar to web resource addressing.
-     
-     - parameter basePath: base string.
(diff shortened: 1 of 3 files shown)


Core/Extensions/Foundation/String.swift
```diff
@@ -20,6 +20,30 @@ public func / (basePath: String, pathComponent: String) -> String {
...
+     
+     - parameter basePath: base string.
+     - parameter pathComponent: natural number that will be appended.
+     
+     - returns: The new NSURL, with pathComponent appended to baseURL.
+ */
+public func / (basePath: String, pathComponent: UInt) -> String {
+    return (basePath as NSString).appendingPathComponent(String(pathComponent))
+}
+
```
FranDepascuali: I think a protocol as a namespace can be used for less repetition.

```swift
public protocol PathAppendable {}

extension Int: PathAppendable {}
extension UInt: PathAppendable {}
extension String: PathAppendable {}

public func / (basePath: String, pathComponent: PathAppendable) -> String {
    return (basePath as NSString).appendingPathComponent(String(describing: pathComponent))
}
```

I think it is okay here to use the `init(describing:)`, as in the docs it says: ```The String(describing:) initializer is the preferred way to convert an instance of any type to a string```

https://developer.apple.com/reference/swift/customstringconvertible

danielaRiesgo: I like it.
Although I'd do something like this:
```swift
public protocol PathAppendable {
    func toString() -> String
}

public extension PathAppendable {
    func toString() -> String {
        return String(describing: self)
    }
}
```

 And glad to hear from you !

woloxci:

Core/Extensions/Foundation/String.swift
```diff
@@ -8,17 +8,7 @@
...
-    
-    - parameter basePath: base string.
-    - parameter pathComponent: string that will be appended.
- 
-    - returns: The new NSURL, with pathComponent appended to baseURL.
- */
-public func / (basePath: String, pathComponent: String) -> String {
-    return (basePath as NSString).appendingPathComponent(pathComponent)
-}
+
```
damfinkel: check the amount of whitespaces here

PR #39, opened 2017-03-22
Completing tests

danielaRiesgo:
## Summary ##
Adding tests for all functionality added since last release and correcting old ones.

### DIFF
--- a/Core/Extensions/Foundation/Date.swift
+++ b/Core/Extensions/Foundation/Date.swift
@@ -25,7 +25,7 @@ public extension Date {
     /**
      Returns a Date with right now's time and date.
      
-     - seealso: NSDate.init()
+     - seealso: Date.init()
     */
     public static func now() -> Date {
         return Date()
@@ -37,7 +37,7 @@ public extension Date {
      - parameter calendar: the calendar to get the date components from. 
      Default: current user calendar.
      
-     - seealso: NSCalendar().component()
+     - seealso: Calendar().component()
      - returns: An Int value representing the hours, between 0 - 24.
      */
     public func hours(using calendar: Calendar = .current) -> Int {
@@ -50,7 +50,7 @@ public extension Date {
      - parameter calendar: the calendar to get the date components from. 
      Default: current user calendar.
      
-     - seealso: NSCalendar().component()
+     - seealso: Calendar().component()
      - returns: An Int value representing the minutes, between 0 - 60.
      */
     public func minutes(using calendar: Calendar = .current) -> Int {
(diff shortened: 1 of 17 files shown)


woloxci:

Core/Extensions/Foundation/String.swift
```diff
@@ -97,18 +97,14 @@ public extension String {
     }
     
     /**
-     Returns true if self contains the given `searchString`, false if not.
-    */
-    public func contains(searchString: String) -> Bool {
```
pablog: Was it added in `Foundation`?

danielaRiesgo: Yes it was.

Core/Extensions/Foundation/Date.swift
```diff
@@ -119,15 +158,10 @@ public extension Date {
...
-        var monthString: String = "\(month)"
-        var dayString: String = "\(day)"
-        if month < 10 {
-            monthString = "0\(month)"
-        }
-        if day < 10 {
-            dayString = "0\(day)"
-        }
-        self.init(dateString: "\(year)-\(monthString)-\(dayString)")
+        let yearString = String(format: "%.4d", year)
```
danielaRiesgo: I run it on the playground and with ".4" I get zeroes before the number, with the "4" I get whitespaces before.

CoreTests/Extensions/Foundation/DictionarySpec.swift
```diff
@@ -0,0 +1,115 @@
...
+public class DictionarySpec: QuickSpec {
+    
+    override public func spec() {
+        
+        describe("#appending(contentsOf:)") {
+            
+            var dict1: [Int: String]!
+            var dict2: [Int: String]!
+            
+            context("when dictionaries have no same keys") {
```
damfinkel: different keys

CoreTests/Extensions/Foundation/URLSpec.swift
```diff
@@ -23,37 +23,34 @@ public class NSURLSpec: QuickSpec {
             let fileManager = FileManager.default
             return fileManager.urls(for: .documentDirectory, in: FileManager.SearchPathDomainMask.userDomainMask).last!
         }()
+        
+        describe("#isHTTPProtocol") {
```
damfinkel: do we want isHTTPProtocol to return true if protocol is HTTPS? Feels wrong if we hace an isProtocolHTTPS method

CoreTests/Extensions/Foundation/StringSpec.swift
```diff
@@ -183,6 +197,197 @@ public class StringSpec: QuickSpec {
             }
         }
         
+        describe("#isNotEmpty") {
```
damfinkel: I know this implementation wasn't done in this PR, but why do we have an "isNotEmpty"? I think it's cleaner to have a "isEmpty" and, in case we want to know the opposite, call !string.isEmpty. Negatives are always confusing

danielaRiesgo: There is always the option to not use it. But I think is more readable `if string.isNotEmpty` than `if !string.isEmpty` (which would be read like `if not (string.isEmpty)` which I think is more confusing or at least takes a little more to think)

CoreTests/Extensions/Foundation/DictionarySpec.swift
danielaRiesgo: Ammending commit

CoreTests/Extensions/Foundation/URLSpec.swift
danielaRiesgo: I get what you mean, but the HTTPS is still an HTTP protocol. Maybe add a new `isOnlyHTTPProtocol` or `isBluntHTTPProtocol` or `isPlainHTTPProtocol`?

CoreTests/Extensions/Foundation/StringSpec.swift
damfinkel: Oh, ok, I didn't know that String already had the isEmpty method. I don't necessarily agree that isNotX is better that !isX, but if the coder can choose, it's fine by me.

CoreTests/Extensions/Foundation/URLSpec.swift
damfinkel: I'm not sure... It's weird to call a method "is OnlyHTTPProtocol"... In any case I don't see the problem of doing `if(x.isHTTPProtocol && !x.isHTTPSProtocol)`. What I'm not sure about is to consider isHTTPProtocol as true if it is an HTTPS protocol. It feels like it's the same as having 1.0.isInteger as true because that particular float is also an integer. But I'm not sure, @pablog what do you think?

pablog: As I was talking to @danielaRiesgo, I'd like to know how these functions would be used or their utility.

Anyway, I'd keep them this way. `URL` can be used with any kind of protocol. Using `isHTTPProtocol` (I assume) can be useful if I need to know if it's whether `HTTP` or `HTTPS`. `isHTTPSProtocol` would be useful to know if the request is secured or not.

PR #40, opened 2017-03-23
Adding dequeue cell functions to table view and collection view

danielaRiesgo:
## Summary ##
Adding dequeue cell functions for UITableView.
Adding all cell functions that UITableView has to UICollectionView.

## Known issues ##
Can't add tests since it requires to load a Nib.

### DIFF
--- a/Core/Extensions/UIKit/UICollectionView.swift
+++ b/Core/Extensions/UIKit/UICollectionView.swift
@@ -1,0 +1,34 @@
+//
+//  UICollectionView.swift
+//  Core
+//
+//  Created by Daniela Riesgo on 3/23/17.
+//  Copyright © 2017 Wolox. All rights reserved.
+//
+
+import UIKit
+
+public extension UICollectionView {
+    
+    /**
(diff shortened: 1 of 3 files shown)


Core/Extensions/UIKit/UICollectionView.swift
```diff
@@ -0,0 +1,34 @@
...
+//
+
+import UIKit
+
+public extension UICollectionView {
+    
+    /**
+         Registers a cell to be used by a UICollectionView.
+         
+         - parameter cellType: An identifiable cell to take the identifier from.
```
pablog: Add parameter bundle

Core/Extensions/UIKit/UICollectionView.swift
```diff
@@ -0,0 +1,34 @@
...
+     */
+    func register(cellType: IdentifiableCell.Type, bundle: Bundle? = .none) {
```
pablog: Add `_` to first argument to make it consistant with `UITableView` extension

woloxci:

PR #46, opened 2017-04-12
Add header footer register in UITableView extension

pablog:
## Summary ##

Add heder and footer register in UITableView extension to do this easily taking advantage of the cell identifier.

## Known issues ##

I've changed the name `register` to `registerCell` to distinguish between the existent ones, and be consistent with the naming. This is not necessarily an issue, but it's not backward compatible. When updating the framework, this code in the project need to be updated too.


Core/Extensions/UIKit/UICollectionView.swift
```diff
@@ -16,8 +16,9 @@ public extension UICollectionView {
          - parameter cellType: An identifiable cell to take the identifier from.
          - parameter bundle: The Bundle where the Nib for the cell is located.
      */
-    public func register(_ cellType: IdentifiableCell.Type, bundle: Bundle? = .none) {
-        register(UINib(nibName: cellType.cellIdentifier, bundle: bundle), forCellWithReuseIdentifier: cellType.cellIdentifier)
+    public func registerCell(_ cellType: IdentifiableCell.Type, bundle: Bundle? = .none) {
```
danielaRiesgo: I think it's better to name it `register(cell cellType: ...)`.

Core/Extensions/UIKit/UITableView.swift
```diff
@@ -16,8 +16,20 @@ public extension UITableView {
          - parameter cellType: An identifiable cell to take the identifier from.
          - parameter bundle: The Bundle where the Nib for the cell is located.
      */
-    public func register(_ cellType: IdentifiableCell.Type, bundle: Bundle? = .none) {
-        register(UINib(nibName: cellType.cellIdentifier, bundle: bundle), forCellReuseIdentifier: cellType.cellIdentifier)
+    public func registerCell(_ cellType: IdentifiableCell.Type, bundle: Bundle? = .none) {
```
danielaRiesgo: Same

Core/Extensions/UIKit/UITableView.swift
```diff
@@ -16,8 +16,20 @@ public extension UITableView {
...
+        register(nib, forCellReuseIdentifier: cellType.cellIdentifier)
+    }
+    
+    /**
+        Registers a header or footer to be used by a UITableView.
+    
+        - parameter cellType: An identifiable cell to take the identifier from.
+        - parameter bundle: The Bundle where the Nib for the cell is located.
+    */
+    func registerHeaderFooter(_ cellType: IdentifiableCell.Type, bundle: Bundle? = .none) {
```
danielaRiesgo: Same about name.
Also, aren't there equivalents for collection view?

danielaRiesgo:
@pablog for travis to work, we should remove the following from the `.travis.yml` file:
```
#For OS 10.2 all simulators are duplicated so we need to delete one
# or a valid destination won't be found. (At least until travis fixes this)
# Deleting an iPhone 6 simulator
- xcrun simctl delete 1FD54EA7-5A25-4D6F-8599-D6F7687DA4EE
```

Core/Extensions/UIKit/UITableView.swift
pablog: There're no equivalentes for `UICollectionView`.

It should be any view conforming `IdentifiableCell`. It can be a `UITableViewCell`.

danielaRiesgo:
@pablog do you think it would be better for backward compatibility to leave the `register(cell:` function as `register(_ cell:` and for all other ones require the label?

pablog:
@danielaRiesgo I prefer this current naming for consistency. It's a trivial fix in case the library is updated. I'd keep this naming.

Core/Extensions/UIKit/UITableView.swift
danielaRiesgo: Yes, but I think header and footers are only needed to be views, not cells, so we are restricting more than necessary. Maybe we should add an IdentifiableView protocol as well.

woloxci:

Core/Extensions/UIKit/UITableView.swift
danielaRiesgo: And for collection views I think there is a register supplementary view as header or footer: https://developer.apple.com/reference/uikit/uicollectionview/1618068-dequeuereusablesupplementaryview

danielaRiesgo:
@pablog I was just thinking, maybe there should be an `IdentifiableCell` and an `IdentifiableView` protocol, just a `NibIdentifiable` protocol for anything that can be loaded from a Nib and has an identifier with the same name.

Because, right now, `IdentifiableCell` is a protocol with a default implementation for cells, but anything could implement it, and we are doing in `dequeue`: `dequeueReusableCell(withIdentifier: cellType.cellIdentifier) as! T` and we can't assure that the `as! T` will always succeed.

I think we should only have a `NibIdentifiable` protocol (or just `Identifiable` since the `dequeue` function just needs an identifier, not a nib) and don't force cast in dequeue functions, just do `as? T`.
So we can give a default implementation to all cells (table view and collection view cells) and also to `UICollectionReusableView`.

woloxci:

danielaRiesgo:
@pablog what do you think of this?

woloxci:

danielaRiesgo:
😄

PR #49, opened 2017-04-20
Adding missing tests and functions along the way

danielaRiesgo:
## Summary ##

Adding missing tests on UI elements like bundle loading nibs.
Having problems testing register and dequeue functions, leaving tests and comments for future fixing. (I think it's because we are using it without a view controller and so the delegate/data source functions are not called correctly in order and so the state is invalid and works weird).
Adding functions in Bundle similar to what was there but more complete.
Making some small corrections to table and collections functions so that the registering is valid (not accept anything that is not the correct type and so it will fail when dequeuing).
For this, reorganizing and creating a global `SimpleName` function to avoid repeating and have more coherence between our protocols and their purpose.
Adding small function in UIViewController related to what was there.


woloxci:

CoreTests/Resources/Resources.swift
```diff
@@ -0,0 +1,49 @@
...
+import Core
+import MBProgressHUD
+
+class NibLoadableCollectionCell: UICollectionViewCell, NibLoadable { }
+class NibLoadableCollectionView: UICollectionReusableView, NibLoadable { }
+
+class NibLoadableTableCell: UITableViewCell, NibLoadable { }
+class NibLoadableTableView: UITableViewHeaderFooterView, NibLoadable {
+
+    static var nibName: String {
```
pablog: Is this necessary? Shouldn't be be inferred from the class name?

CoreTests/Resources/Resources.swift
```diff
@@ -0,0 +1,49 @@
...
+        return "NibLoadableTableViewCustom"
+    }
+
+}
+
+class NibLoadableTableViewCustom: UITableViewHeaderFooterView, NibLoadable {
+    
+    static var nibName: String {
```
pablog: Same

CoreTests/Resources/Resources.swift
```diff
@@ -0,0 +1,49 @@
...
+    
+    static var nibBundle: Bundle {
+        return Bundle(for: MBProgressHUD.self)
+    }
+    
+}
+
+class NibLoadableTableViewCustom2: UITableViewHeaderFooterView, NibLoadable {
+    
+    static var nibName: String {
```
pablog: Same

CoreTests/Resources/Resources.swift
danielaRiesgo: Yes, it's the default implementation, but here I'm changing it on purpose so to see it works for custom names also (see that the ones above don't need to declare anything).

CoreTests/Resources/Resources.swift
danielaRiesgo: This yes, I can remove it.

CoreTests/Resources/Resources.swift
danielaRiesgo: Same as the first. Is a custom name.

Core/Extensions/Foundation/Bundle.swift
```diff
@@ -16,36 +16,156 @@ public extension Bundle {
...
+     - returns: The float associated with key in the receiver's property list (Info.plist)
+            or .none if the key doesn't exist or if the value is not not a real number.
+     - seealso: object(forInfoDictionaryKey:)
+     */
+    public func getFloat(from key: String) -> Float? {
+        guard let value = object(forInfoDictionaryKey: key) as? Float else {
+            return .none
+        }
+        return value
+    }
```
FranDepascuali: You could also add:
```swift
public func get<T>(from key: String) -> T? {
    guard let value = object(forInfoDictionaryKey: key) as? T else {
        return .none
    }
    return value
}
```
or even better
```swift
public func get<T>(from key: String) -> T? {
    return object(forInfoDictionaryKey: key) as? T
}     
```
and avoid repeating the code in each get. 
Maybe just making it public or making it private and calling it from other functions.

Core/Extensions/Foundation/Bundle.swift
```diff
@@ -16,36 +16,156 @@ public extension Bundle {
...
+     
+     - parameter key: A key in the bundle's property list.
+     - returns: The array associated with key in the receiver's property list (Info.plist)
+            or .none if the key doesn't exist or if the value is not an array of elements
+            of type T.
+     - seealso: object(forInfoDictionaryKey:)
+     */
+    public func getArray<T>(from key: String) -> [T]? {
+        guard let value = object(forInfoDictionaryKey: key) as? NSArray,
+              let swiftArray = value as? [T] else {
```
FranDepascuali: it is necessary to convert first to `NSArray` and then to `[T]`? Can't it be directly converted to `[T]`?

Core/Extensions/Foundation/Bundle.swift
```diff
@@ -16,36 +16,156 @@ public extension Bundle {
...
+     
+     - parameter key: A key in the bundle's property list.
+     - returns: The dictionary associated with key in the receiver's property list (Info.plist)
+            or .none if the key doesn't exist or if the value is not a dictionary of
+            keys of type K and values of type V.
+     - seealso: object(forInfoDictionaryKey:)
+     */
+    public func getDictionary<V>(from key: String) -> [String: V]? {
+        guard let value = object(forInfoDictionaryKey: key) as? NSDictionary,
+              let swiftDictionary = value as? [String: V] else {
```
FranDepascuali: it is necessary to convert first to `NSDictionary` and then to `[String: V]`? Can't it be directly converted to `[String: V]`?

Core/Extensions/Foundation/Bundle.swift
FranDepascuali: I think get and subscript should behave the same way. I would propose to have just this two functions:
```swift
public subscript(key: String) -> T? {
    return get(from: key)
}

public func get<T>(from key: String) -> T? {
    return object(forInfoDictionaryKey: key) as? T
}
```
I think it should work with NSArray and NSDictionary too

Core/Extensions/Foundation/Bundle.swift
```diff
@@ -16,36 +16,156 @@ public extension Bundle {
...
+            receiver's property list (Info.plist) or .none if the key doesn't exist.
+     - seealso: object(forInfoDictionaryKey:) and String(describing:)
     */
     public subscript(key: String) -> String? {
-        guard let value = object(forInfoDictionaryKey: key) as? String, !value.isEmpty else {
+        guard let value = object(forInfoDictionaryKey: key) else {
+            return .none
+        }
+        let stringValue = value as? String ?? String(describing: value)
+        return stringValue
```
FranDepascuali: What happens if there is a key and there is no value assigned to it? I remember it caused some problems, so that's why there was a `!value.isEmpty`. 

I think it should continued to be treated as an error (it doesn't make sense to have a key without a value or with a string "" value).

Core/Extensions/Foundation/Bundle.swift
danielaRiesgo: Answering the fist comment:
- The thing is that not all types can be get from an Info.plist, these are the only options, all other will fail. Maybe we could make a PListType protocol and make these types conform it and make the function public for types confirming that protocol, but I don't know if there is a way to avoid someone making other types conform that protocol.

danielaRiesgo: Answering the second comment:

- All this already have tests and are working. Personally I preferred them to work different, because as the subscript can only return one type, I thought of the subscript as something you would always want to work as much as you can and so you want the string representation of the value and parse it or do whatever you want with it. But if you want a specific type, you are expecting that to be that type and if not you most surely would want to know it's not from that type so you change it. That's why I thought of them to work differently, so as to be able to identify when something is not of that type although it can be transformed to it.
@pablog what do you think?

Core/Extensions/Foundation/Bundle.swift
danielaRiesgo: Added tests for that case (if I understood correctly) and they are passing.

woloxci:

:x: Slather: 63.73 - Minimum: 70.0
:white_check_mark: Swiftlint: 98.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/add-tests) for more details.

woloxci:

:x: Slather: 63.73 - Minimum: 70.0
:white_check_mark: Swiftlint: 98.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/add-tests) for more details.

woloxci:

:white_check_mark: Slather: 86.61
:white_check_mark: Swiftlint: 98.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/add-tests) for more details.

woloxci:

:white_check_mark: Code Coverage (Slather): 86.61
:white_check_mark: Slather: 86.61
:white_check_mark: Swiftlint: 98.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/add-tests) for more details.

Core/Extensions/Foundation/Bundle.swift
pablog: I agree with @FranDepascuali, unless we allow empty values, which I'm not fully sure whether we should or not.

I think I'd keep `as? String, !value.isEmpty`. It's the same case the user forgot to add the key and the user forgot to add the value.

Core/Extensions/Foundation/Bundle.swift
pablog: I don't think there's a right answer for this. It depends on what we prefer.

Answering the types: 

I think making the function suggested by @FranDepascuali private and having the current ones as public is the best option. 
 
It's good the code doesn't compile with an invalid type, and I'd like to keep it. But if we can have the same behavior with less lines of code it's better.

Answering the second comment:

I don't think `subscript(key: String) -> String?` is valid in this case, what happens if the user wants to retrieve an array or a dictionary? If we allow that using `getType` we should also this using `subscript`.

danielaRiesgo: Okay, so we were talking with @pablog and got to the conclusion our options are:

- Make the subscript return `Any?`
- Delete subscript and make users use the `getX` functions
- Make the subscript work exactly the same way as the `getString` function, to be used as a shortcut for it.

@damfinkel what do you think?

Core/Extensions/Foundation/Bundle.swift
danielaRiesgo: Ah okay, now I understood. I tried and it only returns empty string if the type is string and it's empty (if it's an empty array or data or etc. it returns none).
So let's think about it. Could a value be purposely empty?
What id they have API info like BASE_URL, SUBDOMAIN, API_VERSION and and somehow at some point the API becomes unversioned or the subdomain isn't needed ? Maybe according to the configuration in the config files, for production they have a versiones api but for development not?

damfinkel: it's not necessarily an error for it to be empty... It may be on purpose (good example that one @danielaRiesgo, even though I can't remember if it ever happened to me). So, the options are:

1) Return .none if it's "". Advantage: 90% of the time, it'll be an error. Disadvantage: if it was intended, we may have to handle something differently for that case
2) Return "" if it's "". Advantage: if it's intended, it works for every case. Disadvantage: it may be an error we are ignoring and it may take debugging time to find it.

I'm leaning towards 1. It's rare that we want to intentionally have a plist property empty, and if that's the case, it may be handled by the consumer of this framework. So I'd go for what @pablog said.

Core/Extensions/Foundation/Bundle.swift
```diff
@@ -16,36 +16,156 @@ public extension Bundle {
...
+     Returns the string associated with the specified key in the receiver's
+     information property list (Info.plist).
+     
+     - parameter key: A key in the bundle's property list.
+     - returns: The string associated with key in the receiver's property list (Info.plist)
+            or .none if the key doesn't exist or if the value is not a string.
+     - seealso: object(forInfoDictionaryKey:)
+     */
+    public func getString(from key: String) -> String? {
+        guard let value = object(forInfoDictionaryKey: key) as? String else {
```
damfinkel: if you decide to change ```subscript```, change this one too

Core/Extensions/Foundation/Bundle.swift
damfinkel: Also, if I have a property with an array, why would I want to return its String description? Has that happened ever?

Core/Extensions/Foundation/Bundle.swift
damfinkel: I don't know why originally we wanted to return String descriptions of stuff in the plist, but I think that subscript should return Any?, it's what instinctively would think it does if I knew a ```subscript``` exists. If it's an array or a dictionary, I'd want that instead of a String representing that. I wouldn't delete subscript. And I'd keep the getInt/getString etc., and use privately what @FranDepascuali suggested

woloxci:

:white_check_mark: Code Coverage (Slather): 85.96
:white_check_mark: Slather: 86.61
:white_check_mark: Swiftlint: 98.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/add-tests) for more details.

woloxci:

:white_check_mark: Code Coverage (Slather): 85.96
:white_check_mark: Slather: 86.61
:white_check_mark: Swiftlint: 98.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/add-tests) for more details.

woloxci:

:white_check_mark: Slather: 86.61
:white_check_mark: Swiftlint: 98.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/add-tests) for more details.

woloxci:

:white_check_mark: Slather: 86.61
:white_check_mark: Swiftlint: 98.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/add-tests) for more details.

Core/Extensions/Foundation/Collection.swift
```diff
@@ -16,5 +16,17 @@ public extension Collection {
...
+
+    /**
+     Returns the first element of the collection for which the predicate returns true,
+     or .none if there is no such element.
+
+     - parameter condition: The condition that will be applied.
+     - returns: The element that satisfies the predicate or .none if none satisfies.
+     - seealso: index(where:)
+     */
+    public func getFirst(where condition: (Iterator.Element) -> Bool) -> Iterator.Element? {
```
pablog: I'm not sure it makes sense:

`Array` already provides a function `filter` which does this.
`Collection` doesn't, because there can be non sorted collections (i.e. `Set`) in which doesn't make sense to ask for `first` since there is no order.

What do you think?

pablog: Cannot it be called `first`?

danielaRiesgo: There already is a `first` property that gives you the first element, there also is a `last` property. Maybe there isn't an obvious order, but there is an order by which you can iterate in all collections.
The difference with array's filter is that filter iterates over all the collection and gives you all elements that satisfies the condition, but sometimes you just need the first, or only one. And so to avoid iterating over the whole collection.

pablog: Yes, It's controversial why there's a `first` in `collection`.

OK, leave it, and call it `first` with a parameter.

woloxci:

:white_check_mark: Slather: 86.61
:white_check_mark: Swiftlint: 98.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/add-tests) for more details.

Issue #50, opened 2017-04-20
Improving view's border functions' name and offset handling


danielaRiesgo:
![](https://github.trello.services/images/mini-trello-icon.png) [Improving view's border functions' name and offset handling](https://trello.com/c/Trc031XW/18-refactorizar-metodos-para-agregar-bordes-de-uiview-para-manejar-mejor-el-tema-de-los-offsets)

UIView's `addTopBorder`, `addLeftBorder`, `addBottomBorder` and `addRightBorder` functions need to have the name improved and handling offset in a clearer and friendly way.

danielaRiesgo:
Solved in PR #51

PR #51, opened 2017-04-21
Improving UIView functions

danielaRiesgo:
## Summary ##

**Issue**: #50 

This was discussed previously in: https://github.com/Wolox/wolmo-core-ios/pull/30#discussion-diff-94456549R37

Improved naming and offset handling.
Code was copied and pasted into Wyz project to try it out.

### DIFF
--- a/Core/Extensions/UIKit/UIView/UIView.swift
+++ b/Core/Extensions/UIKit/UIView/UIView.swift
@@ -25,12 +25,17 @@ public enum ViewPositioning {
 
 extension UIView {
     
-    private func addBorderView(from border: ViewBorder, horizontal: Bool) -> UIView {
+    private enum Direction {
+        case Horizontal
+        case Vertical
+    }
+    
+    private func addBorderView(from border: ViewBorder, direction: Direction) -> UIView {
         let borderView = UIView(frame: .zero)
         borderView.backgroundColor = border.color
         addSubview(borderView)
         borderView.translatesAutoresizingMaskIntoConstraints = false
-        let anchor = horizontal ? borderView.heightAnchor : borderView.widthAnchor
+        let anchor = (direction == .Horizontal) ? borderView.heightAnchor : borderView.widthAnchor
         anchor.constraint(equalToConstant: CGFloat(border.thickness)).isActive = true
         return borderView
     }
@@ -52,7 +57,7 @@ extension UIView {
                     withLeftOffset left: CGFloat = 0, rightOffset right: CGFloat = 0,
                     useConstraints: Bool = true) {
         if useConstraints {
-            let borderView = addBorderView(from: border, horizontal: true)
+            let borderView = addBorderView(from: border, direction: .Horizontal)
             borderView.topAnchor.constraint(equalTo: topAnchor).isActive = true
             borderView.leftAnchor.constraint(equalTo: leftAnchor, constant: left).isActive = true
             rightAnchor.constraint(equalTo: borderView.rightAnchor, constant: right).isActive = true
@@ -83,7 +88,7 @@ extension UIView {
                     withLeftOffset left: CGFloat = 0, rightOffset right: CGFloat = 0,
                     useConstraints: Bool = true) {
         if useConstraints {
-            let borderView = addBorderView(from: border, horizontal: true)
+            let borderView = addBorderView(from: border, direction: .Horizontal)
             borderView.bottomAnchor.constraint(equalTo: bottomAnchor).isActive = true
             borderView.leftAnchor.constraint(equalTo: leftAnchor, constant: left).isActive = true
             rightAnchor.constraint(equalTo: borderView.rightAnchor, constant: right).isActive = true
@@ -114,7 +119,7 @@ extension UIView {
                     withTopOffset top: CGFloat = 0, bottomOffset bottom: CGFloat = 0,
                     useConstraints: Bool = true) {
         if useConstraints {
-            let borderView = addBorderView(from: border, horizontal: false)
+            let borderView = addBorderView(from: border, direction: .Vertical)
             borderView.leftAnchor.constraint(equalTo: leftAnchor).isActive = true
             borderView.topAnchor.constraint(equalTo: topAnchor, constant: top).isActive = true
             bottomAnchor.constraint(equalTo: borderView.bottomAnchor, constant: bottom).isActive = true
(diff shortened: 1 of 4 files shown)


woloxci:

Core/Extensions/UIKit/UIView/UIView.swift
```diff
@@ -25,86 +25,163 @@ public enum ViewPositioning {
...
-        borderView.backgroundColor = border.color
-        addSubview(borderView)
+    public func add(right border: ViewBorder,
+                    withTopOffset top: CGFloat = 0, bottomOffset bottom: CGFloat = 0,
+                    useConstraints: Bool = true) {
+        if useConstraints {
+            let borderView = addBorderView(from: border, horizontal: false)
+            borderView.rightAnchor.constraint(equalTo: rightAnchor).isActive = true
+            borderView.topAnchor.constraint(equalTo: topAnchor, constant: top).isActive = true
+            bottomAnchor.constraint(equalTo: borderView.bottomAnchor, constant: bottom).isActive = true
```
pablog: Is this the same than?

`borderView.bottomAnchor.constraint(equalTo: bottomAnchor, constant: bottom).isActive = true`

If it's the same I'd use this way, so they all look the same. (In every function)

danielaRiesgo: We could, but the constant should be negative then.
I don't have any preference, I can do it in that way.

pablog: No, it's OK then.

Core/Extensions/UIKit/UIView/UIView.swift
```diff
@@ -7,13 +7,13 @@
 //
 import UIKit
 
-public struct Border {
+public struct ViewBorder {
     
-    let height: Float
+    let thickness: Float
```
FranDepascuali: Good good word!

Core/Extensions/UIKit/UIView/UIView.swift
```diff
@@ -25,86 +25,163 @@ public enum ViewPositioning {
 
 extension UIView {
     
+    private func addBorderView(from border: ViewBorder, horizontal: Bool) -> UIView {
```
FranDepascuali: It is correct, but I would find it clearer to have an enum rather than a boolean (though it's the same).
```swift
addBorderView(from: border, direction: .Horizontal)
```
rather than
```swift
addBorderView(from: border, horizontal: true)
```

FranDepascuali: There seems to be something weird about this function signature: I think it would either produce the side-effect (for example, adding the border view) and therefore be named `addBorderView(...) -> Void` or return a new view, named `createBorderView(...)` and return a new border view without injecting the side-effect here and doing it at the caller site.

Core/Extensions/UIKit/UIView/UIView.swift
```diff
@@ -25,86 +25,163 @@ public enum ViewPositioning {
...
-        addSubview(borderView)
+    public func add(top border: ViewBorder,
+                    withLeftOffset left: CGFloat = 0, rightOffset right: CGFloat = 0,
+                    useConstraints: Bool = true) {
+        if useConstraints {
+            let borderView = addBorderView(from: border, horizontal: true)
+            borderView.topAnchor.constraint(equalTo: topAnchor).isActive = true
+            borderView.leftAnchor.constraint(equalTo: leftAnchor, constant: left).isActive = true
+            rightAnchor.constraint(equalTo: borderView.rightAnchor, constant: right).isActive = true
+        } else {
```
FranDepascuali: I think I would not leave this else branch, using the frame for determining the subview has already proven to be a headache (for example, if the parent view gets resized, the subview would continue to be the same size). Although it is true that the default `useConstraints` is true, I would not leave this as it goes against the guidelines.

Core/Extensions/UIKit/UIView/UIView.swift
```diff
@@ -25,86 +25,163 @@ public enum ViewPositioning {
...
-        addSubview(borderView)
+    public func add(bottom border: ViewBorder,
+                    withLeftOffset left: CGFloat = 0, rightOffset right: CGFloat = 0,
+                    useConstraints: Bool = true) {
+        if useConstraints {
+            let borderView = addBorderView(from: border, horizontal: true)
+            borderView.bottomAnchor.constraint(equalTo: bottomAnchor).isActive = true
+            borderView.leftAnchor.constraint(equalTo: leftAnchor, constant: left).isActive = true
+            rightAnchor.constraint(equalTo: borderView.rightAnchor, constant: right).isActive = true
+        } else {
```
FranDepascuali: same here

Core/Extensions/UIKit/UIView/UIView.swift
danielaRiesgo: I get what you mean, but only a create function wouldn't make any sense because it just a view with a background. The add function is useful, but if we don't return anything the user can't take any choice over the view, for example hiding it after something happens.

Core/Extensions/UIKit/UIView/UIView.swift
danielaRiesgo: I would leave it because, if I'm not mistaken but I'm not sure, there are times where you need to use frames, like for cells or things like that. And it's not nice to use frames and have this use constraints.
I would leave the option open to the user, but we can add a note to strongly recommend using constraints or something like that.
@pablog what do you think?

pablog: I'd do what @danielaRiesgo suggests. Adding a comment suggesting this, but letting the user decide how to use it.

Also, I'd be consistent with the other classes in the framework. Do they support both ways?

Core/Extensions/UIKit/UIView/UIView.swift
danielaRiesgo: Now that I think about it, this function is the private one so it doesn't matter, it's to avoid code repetition. But maybe we should add the return in the public functions !

Core/Extensions/UIKit/UIView/UIView.swift
danielaRiesgo: I don't think so. The below one (`public func add(into containerView: UIView, with insets: UIEdgeInsets = .zero, in viewPositioning: ViewPositioning = .front)`) does not give the user the possibility.

woloxci:

:x: Slather: 49.85 - Minimum: 70.0
:white_check_mark: Swiftlint: 98.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/view-border-functions) for more details.

woloxci:

:x: Slather: 49.85 - Minimum: 70.0
:white_check_mark: Swiftlint: 98.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/view-border-functions) for more details.

woloxci:

:x: Slather: 49.85 - Minimum: 70.0
:white_check_mark: Swiftlint: 98.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/view-border-functions) for more details.

Issue #43, opened 2017-04-04
Review `hideBars` var


danielaRiesgo:
`hideBars` var's get in `UIViewcontroller.swift` always returns false.

danielaRiesgo:
![](https://github.trello.services/images/mini-trello-icon.png) [Review hideBars](https://trello.com/c/8selURmp/17-revisar-get-de-var-hidebars-en-uiviewcontroller-swift)

PR #56, opened 2017-05-11
Fixing hideBars getter and improving name.

danielaRiesgo:
## Summary ##
**Issue**: #43 

Correcting `hideBars` getter and improving name to new Swift way.

### DIFF
--- a/Core/Extensions/UIKit/UIViewController.swift
+++ b/Core/Extensions/UIKit/UIViewController.swift
@@ -13,7 +13,7 @@ public extension UIViewController {
     /**
      Toggles the visibility of the status and tabBar.
      
-     This will true only if both bars are hidden.
+     This will be true only if both bars are hidden.
      */
     public var areBarsHidden: Bool {
         set {
--- a/Core/Extensions/UIKit/UIViewController.swift
+++ b/Core/Extensions/UIKit/UIViewController.swift
@@ -11,21 +11,6 @@ import UIKit
 public extension UIViewController {
     
     /**
-     Toggles the visibility of the status and tabBar.
-     
-     This will be true only if both bars are hidden.
-     */
-    public var barsAreHidden: Bool {
-        set {
-            setBarsHidden(newValue)
-        }
-        
-        get {
-            return getBarsHidden()
-        }
-    }
-    
-    /**
      Loads the childViewController into the specified containerView.
      
      It can be done after self's view is initialized, as it uses constraints to determine the childViewController size.
@@ -58,19 +43,6 @@ public extension UIViewController {
         for childController in self.childViewControllers {
             childController.unloadFromParentViewController()
         }
-    }
-    
-}
-
-fileprivate extension UIViewController {
-    
-    fileprivate func setBarsHidden(_ hidden: Bool) {
-        tabBarController?.tabBar.isHidden = hidden
-        UIApplication.shared.isStatusBarHidden = hidden
-    }
-
-    fileprivate func getBarsHidden() -> Bool {
-        return (tabBarController?.tabBar.isHidden ?? true) && UIApplication.shared.isStatusBarHidden
     }
     
 }
--- a/Core/Extensions/UIKit/UIViewController.swift
+++ b/Core/Extensions/UIKit/UIViewController.swift
@@ -12,14 +12,16 @@ public extension UIViewController {
     
     /**
      Toggles the visibility of the status and tabBar.
+     
+     This will true only if both bars are hidden.
      */
-    public var hideBars: Bool {
+    public var areBarsHidden: Bool {
         set {
             setBarsHidden(newValue)
         }
         
         get {
-            return false
+            return getBarsHidden()
         }
     }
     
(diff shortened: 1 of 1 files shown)


Core/Extensions/UIKit/UIViewController.swift
```diff
@@ -12,14 +12,16 @@ public extension UIViewController {
     
     /**
      Toggles the visibility of the status and tabBar.
+     
+     This will true only if both bars are hidden.
```
pablog: will be?

woloxci:

Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/fix-hide-bars) for more details.

Core/Extensions/UIKit/UIViewController.swift
```diff
@@ -60,11 +62,15 @@ public extension UIViewController {
     
 }
 
-private extension UIViewController {
+fileprivate extension UIViewController {
```
pablog: cannot it be private?

Core/Extensions/UIKit/UIViewController.swift
```diff
@@ -60,11 +62,15 @@ public extension UIViewController {
...
     
-    func setBarsHidden(_ hidden: Bool) {
+    fileprivate func setBarsHidden(_ hidden: Bool) {
```
pablog: why do you need it `fileprivate`? cannot you just omit it?

Core/Extensions/UIKit/UIViewController.swift
```diff
@@ -60,11 +62,15 @@ public extension UIViewController {
...
         tabBarController?.tabBar.isHidden = hidden
         UIApplication.shared.isStatusBarHidden = hidden
     }
+
+    fileprivate func getBarsHidden() -> Bool {
```
pablog: same

Core/Extensions/UIKit/UIViewController.swift
danielaRiesgo: Because of the getter and the setter, yes. Or do you see something weird? Or you think I should clarify that if there is no tabBar it won't be taken into account.

pablog: I mean `be` is missing.

Core/Extensions/UIKit/UIViewController.swift
danielaRiesgo: In the new swift a private extension is the same as a fileprivate extension, but I think it's clearer this way.

Core/Extensions/UIKit/UIViewController.swift
danielaRiesgo: I could omit it, but I prefer to have it explicit because many times in the framework we had problems with not specifying modifiers and having, for example, things internal that we wanted public. So I think it's a good practice, at least for a framework, to write every modifier so to be clearer.

woloxci:

Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/fix-hide-bars) for more details.

Core/Extensions/UIKit/UIViewController.swift
pablog: OK, let's leave them then.

Core/Extensions/UIKit/UIViewController.swift
```diff
@@ -12,14 +12,16 @@ public extension UIViewController {
     
     /**
      Toggles the visibility of the status and tabBar.
+     
+     This will be true only if both bars are hidden.
      */
-    public var hideBars: Bool {
+    public var areBarsHidden: Bool {
```
damfinkel: I'm not a fan of the var name, I personally prefer if there is no verb. Something like *barsHidden* is similar to what iOS usually uses (.hidden in UIKit for example). If you must use a verb, I'd rather "barsAreHidden" than "areBarsHidden" which sounds like a question.

danielaRiesgo: I was trying to imitate swift 3's way, since now it uses `isHidden` and `isActive`. But I agree `barsAreHidden` is better. Will change it to that if you agree.

pablog: In swift the [convention](https://swift.org/documentation/api-design-guidelines/#general-conventions) suggests: 

**Uses of Boolean methods and properties should read as assertions about the receiver when the use is nonmutating, e.g. `x.isEmpty`, `line1.intersects(line2)`.**

Anyway, I couldn't find an example where the subject is not `self`. In this case probably `barsAreHidden` may apply better.

@danielaRiesgo what do you think?

danielaRiesgo: Maybe we should use something like `hasBarsHidden` ?

woloxci:

Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/fix-hide-bars) for more details.

Core/Extensions/UIKit/UIViewController.swift
pablog: I'd use `barsAreHidden`, `hasBarsHidden` sounds more like "it can have hidden bars, or have nothing" to someone who doesn't know the background.

Anyway, I'm not fully convinced.

Also, when will it be used? I cannot imagine a case when the user needs to hide status bar and tab bar.

Why not just: 

`tabBar.isHidden`
`tabBar.hide()` 
`statusBar.isHidden`
`statusBar.hide()`

danielaRiesgo: I imagine a use case would be like in a gallery of picture when you want to only see the picture without surroundings, but I'm not completely sure.
 I like the `tabBar.isHidden` and `statusBar.isHidden` since it's more flexible too.

woloxci:

Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/fix-hide-bars) for more details.

woloxci:

Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/fix-hide-bars) for more details.

PR #57, opened 2017-05-11
Making border functions return the view

danielaRiesgo:
## Summary ##

This was discussed before, but it didn't reach a conclusion.

The proposal is to have the border functions return the BorderView so that the user can make changes like hiding it or showing it, or changing color or alpha, or things like that when something in the app happens, like a tap.

The issue that was controversial is that a function creates, adds to self and returns the view. It was discussed to make a function that creates it and another one that adds it. The thing is that the view by itself created is like any other view with nothing, it makes sense when it's added to a view in a certain position, since it's purpose is to serve as a border to something.

I tried to avoid letting the user do things he shouldn't, but there are so many things we can't avoid when giving them the UIView, that I just left a comment in the documentation.

### DIFF
--- a/Core/Extensions/UIKit/UIView/UIView.swift
+++ b/Core/Extensions/UIKit/UIView/UIView.swift
@@ -7,43 +7,96 @@
 //
 import UIKit
 
-public struct ViewBorder {
+/**
+    Properties a BorderView has.
+ */
+public struct BorderViewProperties {
     
-    let thickness: Float
-    let color: UIColor
+    public let thickness: Float
+    public let color: UIColor
     
-    init(thickness: Float, color: UIColor) {
+    public init(thickness: Float, color: UIColor) {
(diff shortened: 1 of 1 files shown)


woloxci:

Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/improving-uiview-functions) for more details.

danielaRiesgo:
Yes, we could. Maybe we should start using it and see what is expected from it so that we start adding it to that wrapper.
But what if the user wants to constraint something to the border, so that it doesn't appear behind it? Should we expose all anchors? But then the user can still do weird things with it.

woloxci:

Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/improving-uiview-functions) for more details.

PR #59, opened 2017-05-12
Added function in button for setting title and image.

danielaRiesgo:
## Summary ##

Adding function for setting title and image on UIButton. Extracted from Wyz, where it worked as expected.

### DIFF
--- a/Core/Extensions/UIKit/UIButton.swift
+++ b/Core/Extensions/UIKit/UIButton.swift
@@ -30,5 +30,41 @@ public extension UIButton {
         let underlinedText = NSAttributedString(string: title, attributes: attributes)
         setAttributedTitle(underlinedText, for: state)
     }
+
+    /* 
+     Sets the button's title and image with the specified spacing and arrangement.
+    
+     - parameter title: Title for the button.
+     - parameter image: Image for the button.
+     - parameter spacing: Spacing between the title and the button.
+     - parameter imageOnRight: Boolean indicating whether the the image should be to the
+            right of the text or to the left. By default, false (to the left).
+     
+     - seealso: CGAffineTransform(scaleX:y:)
+    **/
+    public func setTitle(title: String, image: UIImage, spacing: CGFloat = 0, imageOnRight: Bool = false) {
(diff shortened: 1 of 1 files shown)


Core/Extensions/UIKit/UIButton.swift
```diff
@@ -30,5 +30,41 @@ public extension UIButton {
...
+    
+     - parameter title: Title for the button.
+     - parameter image: Image for the button.
+     - parameter spacing: Spacing between the title and the button.
+     - parameter imageOnRight: Boolean indicating whether the the image should be to the
+            right of the text or to the left. By default, false (to the left).
+     
+     - seealso: CGAffineTransform(scaleX:y:)
+    **/
+    public func setTitle(title: String, image: UIImage, spacing: CGFloat = 0, imageOnRight: Bool = false) {
```
pablog: I'd use an `enum: .left, .right` instead of a `Bool`. But I know most of the framework is done this way.

Probably we can add an issue to refactor all of them, I think it is semantically more accurate.

woloxci:

Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/button-with-image) for more details.

Core/Extensions/UIKit/UIButton.swift
danielaRiesgo: Issue added: https://github.com/Wolox/wolmo-core-ios/issues/60
What other case do you think should go into that? For example, `useConstraints` should be an enum which is either `.frame` or `.constraint` ?

I can't think of another. Since the other boolean left is `animated` in collapsed, but that imitates apple's way.

pablog: `animated` is OK this way since it can be animated or nothing.

`useConstraints` is different given it can use constraints or frame. Probably we can have an `enum` for this too (`LayoutMode`?).

woloxci:

Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/button-with-image) for more details.

PR #65, opened 2017-06-01
Adding default value functions to Dictionary

danielaRiesgo:
## Summary ##

Adding a get to dictionaries where you can provide a default value.
Taken from [this](https://www.swiftbysundell.com/posts/using-autoclosure-when-designing-swift-apis?utm_campaign=This%2BWeek%2Bin%2BSwift&utm_medium=email&utm_source=This_Week_in_Swift_133) post.

## Known issues

For dictionaries whose Value is exactly Any (== Any, not : Any), it seems useful to be able to cast it directly (as the post proposes), but we can't do it with the same name as the function for all dictionaries because they collide and finally apple interprets it as the Value function and ignores the generic function.
I couldn't think of a better thing to do than changing the name for them not to collide, because I couldn't find a way to create for example an extension for all dictionaries whose value is not Any.

### DIFF
--- a/Core/Extensions/Foundation/Dictionary.swift
+++ b/Core/Extensions/Foundation/Dictionary.swift
@@ -52,9 +52,15 @@ public extension Dictionary where Value == Any {
          if the key didn't have a value or if it couldn't be casted to T.
 
          Uses autoclosure so your default value won't be created if it's not needed.
+     
+         - warning: This will raise a runtime error if there is a value in the dictionary
+                but it's not of the same time as the default value. For that case, you should
+                handle it some other way.
      */
     public func castedValue<T>(forKey key: Key, or defaultValue: @autoclosure () -> T) -> T {
-       return self[key] as? T ?? defaultValue()
+        guard let value = self[key] else { return defaultValue() }
+        //swiftlint:disable:next force_cast
+        return value as! T
     }
 
 }
--- a/Core/Extensions/Foundation/Dictionary.swift
+++ b/Core/Extensions/Foundation/Dictionary.swift
@@ -31,5 +31,30 @@ public extension Dictionary {
             updateValue(value, forKey: key)
         }
     }
-    
+
+    /**
+         Returns the value associated with the specified key,
+         or the specified default value if the key didn't have a value.
+     
+         Uses autoclosure so your default value won't be created if it's not needed.
+     */
+    public func value(forKey key: Key, or defaultValue: @autoclosure () -> Value) -> Value {
+        return self[key] ?? defaultValue()
+    }
+
+}
+
+public extension Dictionary where Value == Any {
+
+    /**
+         Returns the value associated with the specified key casted to T
+         (the type of the default value), or the specified default value
+         if the key didn't have a value or if it couldn't be casted to T.
+
+         Uses autoclosure so your default value won't be created if it's not needed.
+     */
+    public func castedValue<T>(forKey key: Key, or defaultValue: @autoclosure () -> T) -> T {
+       return self[key] as? T ?? defaultValue()
+    }
+
 }
--- a/Core/Extensions/Foundation/Dictionary.swift
+++ b/Core/Extensions/Foundation/Dictionary.swift
@@ -53,14 +53,16 @@ public extension Dictionary where Value == Any {
 
          Uses autoclosure so your default value won't be created if it's not needed.
      
-         - warning: This will raise a runtime error if there is a value in the dictionary
+         - warning: This will raise a runtime assertion error if there is a value in the dictionary
                 but it's not of the same time as the default value. For that case, you should
                 handle it some other way.
      */
     public func castedValue<T>(forKey key: Key, or defaultValue: @autoclosure () -> T) -> T {
         guard let value = self[key] else { return defaultValue() }
-        //swiftlint:disable:next force_cast
-        return value as! T
+        guard let castedValue = value as? T else {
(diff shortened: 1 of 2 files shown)


woloxci:

:white_check_mark: Code Coverage (Slather): 87.35


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/add-dictionary-get-with-default) for more details.

CoreTests/Extensions/Foundation/DictionarySpec.swift
```diff
@@ -109,6 +109,75 @@ public class DictionarySpec: QuickSpec {
...
+            beforeEach {
+                dict1 = [1: "one", 2: 2, 3: [",", ".", ":"]]
+            }
+
+            context("when the key is already in the dictionary") {
+
+                context("when the key's value's type is not T") {
+
+                    it("should return the default value") {
+                        let value = dict1.castedValue(forKey: 2, or: "two")
```
damfinkel: So this castedValue tries to return the value for the key, but if it isn't of type T (same type as the default value) it returns the default value? I feel like it's not very intuitive, but I have no idea how to make it more intuitive without explicitly using a class, so let's keep it this way for now

Core/Extensions/Foundation/Dictionary.swift
```diff
@@ -31,5 +31,30 @@ public extension Dictionary {
...
+
+    /**
+         Returns the value associated with the specified key casted to T
+         (the type of the default value), or the specified default value
+         if the key didn't have a value or if it couldn't be casted to T.
+
+         Uses autoclosure so your default value won't be created if it's not needed.
+     */
+    public func castedValue<T>(forKey key: Key, or defaultValue: @autoclosure () -> T) -> T {
+       return self[key] as? T ?? defaultValue()
```
pablog: I'd not do this. In my opinion of types don't match is due to a programming error and should crash, otherwise I think we are "hiding" an error. Can you imagine a use case when this is useful?

danielaRiesgo: Personally, I can't. It was in the post I said in the summary, I guessed it should be useful for some case but I couldn't think of any.  I agree that if it's not that type, it most probably means an error.
But if you just wanted the default value if there isn't a value to the key, which I think could be useful, you'd still need to do this to cast it because you can't return Any or T. We could use `as!` but that isn't a catchable error by the user.

damfinkel: I think this may be useful in cases like WYZ were the api returns non-typed dictionaries (I'm pretty sure I've seen something like that in WYZ). In those cases, if for some reason we want to use that dictionary instead of creating a model (which I think it's usually better), we may want to do for example:

```
let dictionary = ["email":"something@wolox.com.ar", "age":20] // user missing optional bio
let bio = dictionary.castedValue("bio", or: ""); // if we want "" instead of nil
```
instead of doing:
```
let dictionary = ["email":"something@wolox.com.ar", "age":20] // user missing optional bio
let bio = dictionary["bio"] as? T ?? ""
```

But the second option also makes it more explicit, and at some point you may want to change it so you do something if the bio's not there, or throw an exception. 
So I think that, given that castedValue() kind of surpasses swift's type system, it might not me the best practise.

danielaRiesgo: If you want to check if bio exists or not, you still can, but this function may not be useful for you.
The thing is that if bio doesn't exists, you may surely want a default, but what if bio is there and it's an int. Wouldn't you want to know that API is not responding something that it should? Because you can't distinguish those cases with this function.

pablog: I still think I'd not use `?`. Instead, I'd use `!`, so the user is responsible for making the types match.

danielaRiesgo: @damfinkel @nahuelwolox  what do you say?

nahuelwolox: I agree that we shouldn't do this, the only case where I see it as useful is in cases where you know that something is wrong and you want to ignore it (as in that example when API is returning a bad type)... but I don't think we should give that functionality because basically is making it easy for the user to do something wrong and ignore it. If he or she wants to do something like this for some reason, I would prefer that the user is responsible for that and has to explain why it needs it instead of using a method form our own library by default and hiding possible errors

woloxci:

:white_check_mark: Code Coverage (Slather): 87.4


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/add-dictionary-get-with-default) for more details.

woloxci:

:white_check_mark: Code Coverage (Slather): 87.45


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/add-dictionary-get-with-default) for more details.

danielaRiesgo:
@pablog @damfinkel please check the latest changes

Core/Extensions/Foundation/Dictionary.swift
damfinkel: @danielaRiesgo I agree with Nahue

woloxci:

:white_check_mark: Code Coverage (Slather): 87.45


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/add-dictionary-get-with-default) for more details.

Issue #67, opened 2017-06-08
Add UIViewController extensions to modify the nav bar


nahuelwolox:
https://trello.com/c/kKDoytV9

danielaRiesgo:
Solved in PR #68 .

PR #68, opened 2017-06-08
Add UIViewController extension to configure navigation bar

nahuelwolox:
## Summary ##
**Issue**: https://github.com/Wolox/wolmo-core-ios/issues/67

Added an extension of UIViewController to be able to configure several aspects of the navigation bar: Adding a back button, changing the bar tint color, setting right and left buttons and using a label as the title.

### DIFF
--- a/Core/Extensions/UIKit/UIViewController.swift
+++ b/Core/Extensions/UIKit/UIViewController.swift
@@ -47,19 +47,19 @@ public extension UIViewController {
     
 }
 
-//MARK: - Navigation Bar
+// MARK: - Navigation Bar
 
 public extension UIViewController {
 
     /**
-     Configures the navigation bar to have a back button with a particular image.
-     - parameter backImage: The image of the back button.
+     Configures the navigation bar to have a particular image as back button.
+     - parameter image: The image of the back button.
      - warning: This function must be called when self is inside a navigation controller. If not it will arise a runtime nil-force-unwrapping error.
      */
-    func configureBasicNavigationBarWithBackButton(backImage: UIImage) {
+    public func setNavigationBarBackButton(image: UIImage) {
         navigationController!.navigationBar.topItem?.title = ""
-        navigationController!.navigationBar.backIndicatorImage = backImage
-        navigationController!.navigationBar.backIndicatorTransitionMaskImage = backImage
+        navigationController!.navigationBar.backIndicatorImage = image
+        navigationController!.navigationBar.backIndicatorTransitionMaskImage = image
     }
     
     /**
@@ -67,7 +67,7 @@ public extension UIViewController {
      - parameter color: The new color of the navigation bar.
      - warning: This function must be called when self is inside a navigation controller. If not it will arise a runtime nil-force-unwrapping error.
      */
-    func configureNavigationBarColor(_ color: UIColor) {
+    public func setNavigationBarColor(_ color: UIColor) {
         navigationController!.navigationBar.barTintColor = color
     }
     
@@ -75,7 +75,7 @@ public extension UIViewController {
      Sets a collection of buttons as the navigation bar left buttons.
      - parameter buttons: the Array of buttons to use.
      */
-    func configureNavigationLeftButtons(_ buttons: [UIBarButtonItem]) {
+    public func setNavigationLeftButtons(_ buttons: [UIBarButtonItem]) {
         navigationItem.leftBarButtonItems = buttons
     }
     
@@ -83,17 +83,17 @@ public extension UIViewController {
      Sets a collection of buttons as the navigation bar right buttons.
      - parameter buttons: the Array of buttons to use.
      */
-    func configureNavigationRightButtons(_ buttons: [UIBarButtonItem]) {
+    public func setNavigationRightButtons(_ buttons: [UIBarButtonItem]) {
         navigationItem.rightBarButtonItems = buttons
     }
 
     /**
      Adds and configures a label to use as title of the navigation bar.
      - parameter title: the string of the label.
-     - parameter font: the font to use for the label. Default: .systemFont size 18
-     - parameter color: the color of the text. Default: .white
+     - parameter font: the font to use for the label.
+     - parameter color: the color of the text.
      */
-    func setNavBarTitle(_ title: String, font: UIFont, color: UIColor) {
+    public func setNavigationBarTitle(_ title: String, font: UIFont, color: UIColor) {
         let label = UILabel(frame: .zero)
         label.backgroundColor = .clear
         label.font = font
--- a/Core/Extensions/UIKit/UIViewController.swift
+++ b/Core/Extensions/UIKit/UIViewController.swift
@@ -54,7 +54,8 @@ public extension UIViewController {
     /**
      Configures the navigation bar to have a particular image as back button.
      - parameter image: The image of the back button.
-     - warning: This function must be called when self is inside a navigation controller. If not it will arise a runtime nil-force-unwrapping error.
+     - warning: This function must be called when self is inside a navigation controller. 
+            If not it will arise a runtime nil-force-unwrapping error.
      */
     public func setNavigationBarBackButton(image: UIImage) {
         navigationController!.navigationBar.topItem?.title = ""
@@ -65,7 +66,8 @@ public extension UIViewController {
     /**
      Configures the navigation bar color.
      - parameter color: The new color of the navigation bar.
-     - warning: This function must be called when self is inside a navigation controller. If not it will arise a runtime nil-force-unwrapping error.
+     - warning: This function must be called when self is inside a navigation controller. 
+            If not it will arise a runtime nil-force-unwrapping error.
      */
     public func setNavigationBarColor(_ color: UIColor) {
         navigationController!.navigationBar.barTintColor = color
--- a/Core/Extensions/UIKit/UIViewController.swift
+++ b/Core/Extensions/UIKit/UIViewController.swift
@@ -110,4 +110,12 @@ public extension UIViewController {
         navigationItem.titleView = label
     }
     
+    /**
+     Adds an ImageView with the image passed as the titleView of the navigation bar.
+     - parameter image: The image to set as title.
+     */
+    public func setNavigationBarTitleImage(_ image: UIImage) {
+        navigationItem.titleView = UIImageView(image: image)
+    }
+    
 }
--- a/Core/Extensions/UIKit/UIViewController.swift
+++ b/Core/Extensions/UIKit/UIViewController.swift
@@ -47,11 +47,14 @@ public extension UIViewController {
     
 }
 
+//MARK: - Navigation Bar
+
 public extension UIViewController {
 
     /**
-     Configures the navigation bar to have a back button.
+     Configures the navigation bar to have a back button with a particular image.
      - parameter backImage: The image of the back button.
+     - warning: This function must be called when self is inside a navigation controller. If not it will arise a runtime nil-force-unwrapping error.
      */
     func configureBasicNavigationBarWithBackButton(backImage: UIImage) {
         navigationController!.navigationBar.topItem?.title = ""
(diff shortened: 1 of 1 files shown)


Core/Extensions/UIKit/UIViewController.swift
```diff
@@ -46,3 +46,58 @@ public extension UIViewController {
...
+        navigationItem.rightBarButtonItems = buttons
+    }
+
+    /**
+     Adds and configures a label to use as title of the navigation bar.
+     - parameter title: the string of the label.
+     - parameter font: the font to use for the label. Default: .systemFont size 18
+     - parameter color: the color of the text. Default: .white
+     */
+    func setNavBarTitle(_ title: String, font: UIFont = .systemFont(ofSize: 18), color: UIColor = .white) {
```
pablog: I'd remove the default for the font

woloxci:

:white_check_mark: Code Coverage (Slather): 87.2


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/navbar-extensions) for more details.

Core/Extensions/UIKit/UIViewController.swift
```diff
@@ -46,3 +46,58 @@ public extension UIViewController {
...
 }
+
+public extension UIViewController {
+
+    /**
+     Configures the navigation bar to have a back button.
+     - parameter backImage: The image of the back button.
+     */
+    func configureBasicNavigationBarWithBackButton(backImage: UIImage) {
+        navigationController!.navigationBar.topItem?.title = ""
```
danielaRiesgo: What is this for? If before calling this I set a title to the nav bar, does this clean it? or it just cleans the back button's title?

Core/Extensions/UIKit/UIViewController.swift
```diff
@@ -46,3 +46,58 @@ public extension UIViewController {
     }
     
 }
+
+public extension UIViewController {
```
danielaRiesgo: You could put a `//MARK: - Navigation Bar`.

Core/Extensions/UIKit/UIViewController.swift
danielaRiesgo: And for the color.

danielaRiesgo: Why directly setting the title to the nav controller isn't enough?

Core/Extensions/UIKit/UIViewController.swift
```diff
@@ -46,3 +46,58 @@ public extension UIViewController {
...
+
+    /**
+     Configures the navigation bar to have a back button.
+     - parameter backImage: The image of the back button.
```
danielaRiesgo: In all functions that unwraps the navigation controller I would add:
`- warning: This function must be call when self is inside a navigation controller. If not it will arise a runtime nil-force-unwrapping error.`
or something like that.

Core/Extensions/UIKit/UIViewController.swift
```diff
@@ -46,3 +46,58 @@ public extension UIViewController {
     }
     
 }
+
+public extension UIViewController {
+
+    /**
+     Configures the navigation bar to have a back button.
```
danielaRiesgo: `... a back button with a particular image`.

Core/Extensions/UIKit/UIViewController.swift
nahuelwolox: No, the back button by default has a "Back" title atached. This line clears that so that you only see the image.

Core/Extensions/UIKit/UIViewController.swift
nahuelwolox: You can set the title directly to `navigationItem.title` and then set an array of attributes in `navigationController.navigationBar.titleTextAttributes`... But using a label allows us to use `adjustFontSizeToFitWidth` and `.sizeToFit()`, so the title dinamically resizes.

woloxci:

:white_check_mark: Code Coverage (Slather): 87.2


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/navbar-extensions) for more details.

Core/Extensions/UIKit/UIViewController.swift
```diff
@@ -46,3 +46,62 @@ public extension UIViewController {
...
+//MARK: - Navigation Bar
+
+public extension UIViewController {
+
+    /**
+     Configures the navigation bar to have a back button with a particular image.
+     - parameter backImage: The image of the back button.
+     - warning: This function must be called when self is inside a navigation controller. If not it will arise a runtime nil-force-unwrapping error.
+     */
+    func configureBasicNavigationBarWithBackButton(backImage: UIImage) {
```
FranDepascuali: This should be public, as it is meant to be used outside the framework (without modifier it is internal and only usable inside the framework)

Core/Extensions/UIKit/UIViewController.swift
```diff
@@ -46,3 +46,62 @@ public extension UIViewController {
...
+        navigationController!.navigationBar.topItem?.title = ""
+        navigationController!.navigationBar.backIndicatorImage = backImage
+        navigationController!.navigationBar.backIndicatorTransitionMaskImage = backImage
```
FranDepascuali: It is not good to crash from within the framework. These ! should be correctly unwrapped

Core/Extensions/UIKit/UIViewController.swift
```diff
@@ -46,3 +46,62 @@ public extension UIViewController {
...
+    }
+    
+    /**
+     Configures the navigation bar color.
+     - parameter color: The new color of the navigation bar.
+     - warning: This function must be called when self is inside a navigation controller. If not it will arise a runtime nil-force-unwrapping error.
+     */
+    func configureNavigationBarColor(_ color: UIColor) {
+        navigationController!.navigationBar.barTintColor = color
```
FranDepascuali: Add public modifier to function.
It is not good to crash from within the framework. These ! should be correctly unwrapped.

Core/Extensions/UIKit/UIViewController.swift
```diff
@@ -46,3 +46,62 @@ public extension UIViewController {
...
+    }
+    
+    /**
+     Sets a collection of buttons as the navigation bar left buttons.
+     - parameter buttons: the Array of buttons to use.
+     */
+    func configureNavigationLeftButtons(_ buttons: [UIBarButtonItem]) {
```
FranDepascuali: public

Core/Extensions/UIKit/UIViewController.swift
```diff
@@ -46,3 +46,62 @@ public extension UIViewController {
...
+        navigationItem.leftBarButtonItems = buttons
+    }
+    
+    /**
+     Sets a collection of buttons as the navigation bar right buttons.
+     - parameter buttons: the Array of buttons to use.
+     */
+    func configureNavigationRightButtons(_ buttons: [UIBarButtonItem]) {
```
FranDepascuali: public

Core/Extensions/UIKit/UIViewController.swift
```diff
@@ -46,3 +46,62 @@ public extension UIViewController {
...
+        navigationItem.rightBarButtonItems = buttons
+    }
+
+    /**
+     Adds and configures a label to use as title of the navigation bar.
+     - parameter title: the string of the label.
+     - parameter font: the font to use for the label. Default: .systemFont size 18
+     - parameter color: the color of the text. Default: .white
+     */
+    func setNavBarTitle(_ title: String, font: UIFont, color: UIColor) {
```
FranDepascuali: Add public modifier. Also, you should not abbreviate navigation to nav.

Core/Extensions/UIKit/UIViewController.swift
FranDepascuali: I don't particularly like the "Basic" word, as it ambiguous.  If the idea is to change the back button image, just do that. According to the description `Configures the navigation bar to have a back button with a particular image.`, changing the title is a side effect and probably not intended.

Core/Extensions/UIKit/UIViewController.swift
FranDepascuali: I would prefer the prefix `set` rather than `configure`. Two reasons: `Sets a collection of buttons as the navigation bar right buttons.` and configure implies that there is something else apart from setting a value, which in this case isn't like that.

Core/Extensions/UIKit/UIViewController.swift
FranDepascuali: It says that is has default values but in the function definition there aren't default ones.

Core/Extensions/UIKit/UIViewController.swift
danielaRiesgo: So even if you pass a font of size 18 it may show smaller because it didn't enter?
I would add it to documentation, how it works. And what can you use instead if not.

Core/Extensions/UIKit/UIViewController.swift
danielaRiesgo: I agree. It is also no needed. So you could just call it `configureBackButton(with image: UIImage)`. And maybe change documentation to `Configures the navigation bar to have a particular image as back button.`. So that it's implied that it will be only the image and no title.

Core/Extensions/UIKit/UIViewController.swift
```diff
@@ -62,6 +65,7 @@ public extension UIViewController {
     /**
      Configures the navigation bar color.
      - parameter color: The new color of the navigation bar.
+     - warning: This function must be called when self is inside a navigation controller. If not it will arise a runtime nil-force-unwrapping error.
      */
     func configureNavigationBarColor(_ color: UIColor) {
```
danielaRiesgo: I would choose to call everything `set` or everything `configure`, but not have both.

Core/Extensions/UIKit/UIViewController.swift
danielaRiesgo: Agree with public modifier. About the other issue, I agree it's not best for a framework to crash, but why would you call this function if you don't expect a navigation controller to be there? Having it crash may be a way of realizing that you are doing something wrong.

nahuelwolox: Maybe we could print an warning if there's no `navigationController`, but I agree that we shouldn't silently avoid the error by doing nothing if the reference is `null`. In this case I think crashing it's ok because the error the user will see is self explanatory and the precondition is not something weird or arbitrary.

woloxci:

:white_check_mark: Code Coverage (Slather): 87.2


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/navbar-extensions) for more details.

Core/Extensions/UIKit/UIViewController.swift
danielaRiesgo: Maybe instead of crashing with "Error at unwrapping nil value" have a fatalError with our own message with "There is no navigation controller to configure." (or something like that).

danielaRiesgo: This is still pending.

woloxci:

:white_check_mark: Code Coverage (Slather): 87.2


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/navbar-extensions) for more details.

woloxci:

:white_check_mark: Code Coverage (Slather): 87.2


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/navbar-extensions) for more details.

pablog:
I mean, the function should receive the image as parameter.

woloxci:

:white_check_mark: Code Coverage (Slather): 87.2


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/navbar-extensions) for more details.

woloxci:

:white_check_mark: Code Coverage (Slather): 87.2


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/navbar-extensions) for more details.

Issue #69, opened 2017-06-09
Add UIImage extension for image resizing 


nahuelwolox:
https://trello.com/c/fEV8kVgJ

damfinkel:
https://github.com/Wolox/wolmo-core-ios/pull/70

PR #70, opened 2017-06-09
Add UIImage extension for resizing and corner radius

nahuelwolox:
## Summary ##
**Issue**: https://github.com/Wolox/wolmo-core-ios/issues/69

Added an extension to UIImage to resize images maintaining the aspect ratio of the original image, and also add cornerRadius if needed.
This is useful when fetching an image that needs to be shown in a smaller size, or when you need to show that image as a marker in some kind of map so you need to make it smaller and also add corner radius.
Also it could be used to send smaller images to API.

I tested the visual results in another app (with the image view set to `Center` so it shows the actual size of the image regardless of the size of the view):

## Screenshots ##

**Normal Size**
normal size

**Resized to 200x200**
200x200

**Resized to 200x200 with a Corner Radius of 25**
200x200 25cr

**Resized to 320x430 with a Corner Radius of 50**
320x430 50cr

## Known Issues ##
This extension maintains the aspect ratio of the original image. We could easily add a boolean value to determine if we should maintain it or allow for deformation, I didn't expose that because I don't see an utility for it right now but let me know. Also, the corner radius impact will depend on the final size of the image, the same corner radius can give different results on different sized images. Maybe we could add an enum like`CornerRadiusType { .Weak, .Medium, .Strong, .FullCircle }` and calculate the value. 
Also you may not want the corner radius to originate from `(0,0)` and cover the whole image, that's easily exposable too, depends on how much customization we want to offer.


woloxci:

:x: Code Coverage (Slather): 83.3 - Minimum: 85.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/image-extension) for more details.

Core/Extensions/UIKit/UIImage.swift
```diff
@@ -0,0 +1,42 @@
...
+//
+//  Created by Nahuel Gladstein on 6/8/17.
+//  Copyright © 2017 Wolox. All rights reserved.
+//
+
+import UIKit
+
+public extension UIImage {
+    
+    func resizedImage(_ wantedSize: CGSize, cornerRadius: CGFloat = 0.0) -> UIImage {
```
FranDepascuali: This should be public, as it is meant to be used outside the framework (without modifier it is internal and only usable inside the framework)

Core/Extensions/UIKit/UIImage.swift
```diff
@@ -0,0 +1,42 @@
...
+        UIGraphicsEndImageContext()
+        
+        return newImage
+    }
+
+}
+
+private extension CGSize {
+
+    func resizedMaintainingRatio(wantedSize: CGSize) -> CGSize {
```
FranDepascuali: same

Core/Extensions/UIKit/UIImage.swift
danielaRiesgo: I would use another name, maybe like `resizedTo(_ wantedSize: cornerRadius:)`.

Core/Extensions/UIKit/UIImage.swift
nahuelwolox: this is not meant to be exposed unless you think it's worth exposing, I just used it for convenience

Core/Extensions/UIKit/UIImage.swift
nahuelwolox: doesn't swift uses the access modifier of the extension? in that case as the extension is `public` every method without an explicit access modifier would be `public` also

Core/Extensions/UIKit/UIImage.swift
danielaRiesgo: If it's not intended then set extension and function to `fileprivate` as it's done in the rest of the framework.
I don't see much use for the function although maybe for someone that works with frame it may be useful.

nahuelwolox: you mean changing `private` to `fileprivate`? for consistency or is there any other reason?

Core/Extensions/UIKit/UIImage.swift
danielaRiesgo: Yes and no exactly. In this case, yes, `public` default to `public`, but we decided time ago that for frameworks we would do everything completely explicit to be conscious of what we are doing. And because it doesn't always work the same. for example, a `private` extension means its functions are `fileprivate`. So to avoid this confusion and always be clear we use `fileprivate` for extension and always add the access modifiers even if it's redundant.

woloxci:

:x: Code Coverage (Slather): 83.3 - Minimum: 85.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/image-extension) for more details.

Core/Extensions/UIKit/UIImage.swift
```diff
@@ -0,0 +1,42 @@
...
+//
+//  Created by Nahuel Gladstein on 6/8/17.
+//  Copyright © 2017 Wolox. All rights reserved.
+//
+
+import UIKit
+
+public extension UIImage {
+    
+    public func resizedTo(_ wantedSize: CGSize, cornerRadius: CGFloat = 0.0) -> UIImage {
```
pablog: Don't forget to add documentation for the public functions

danielaRiesgo: And a little comment to the private ones if you like ...

woloxci:

:x: Code Coverage (Slather): 83.3 - Minimum: 85.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/image-extension) for more details.

CoreTests/Extensions/UIKit/UIImageSpec.swift
```diff
@@ -0,0 +1,82 @@
...
+//  Copyright © 2017 Wolox. All rights reserved.
+//
+
+import Quick
+import Nimble
+import Core
+
+public class UIImageSpec: QuickSpec {
+    
+    public func ratio(of image: UIImage) -> CGFloat {
```
danielaRiesgo: Maybe this could go to UIImage extensions.

CoreTests/Extensions/UIKit/UIImageSpec.swift
```diff
@@ -0,0 +1,82 @@
...
+                    }
+                }
+                
+                context("And has a different aspect ratio") {
+                    
+                    it("should return an image with the same aspect ratio and a size close to the wanted one") {
+                        let wantedSize = CGSize(width: 250, height: 346)
+                        let resizedImage = originalImage.resizedTo(wantedSize)
+                        expect(self.ratio(of: originalImage)).to(equal(self.ratio(of: resizedImage)))
+                        expect(self.hasMatchingHeightOrWidth(sizeOne: resizedImage.size, sizeTwo: wantedSize)).to(beTrue())
```
danielaRiesgo: I would do it "by hand": 
``` 
let expectedSize = CGSize(width: X, height: Y)
...
expect(resizedImage.size).to(equal(expectedSize))
```

CoreTests/Extensions/UIKit/UIImageSpec.swift
```diff
@@ -0,0 +1,82 @@
...
+                    }
+                }
+                
+                context("And has a different aspect ratio") {
+                    
+                    it("should return an image with the same aspect ratio and a size close to the wanted one") {
+                        let wantedSize = CGSize(width: 15, height: 29)
+                        let resizedImage = originalImage.resizedTo(wantedSize)
+                        expect(self.ratio(of: originalImage)).to(equal(self.ratio(of: resizedImage)))
+                        expect(self.hasMatchingHeightOrWidth(sizeOne: resizedImage.size, sizeTwo: wantedSize)).to(beTrue())
```
danielaRiesgo: Same here.

CoreTests/Extensions/UIKit/UIImageSpec.swift
nahuelwolox: The problem with that is that I don't have an expected size really, because it depends on the aspect ratio of both images and the multiplication factor that arises from that difference (unless the aspect ratio is the same on both images). The resized image will try to fit roughly in the wanted size but I don't really know how unless I test the implementation, and that's why I can only test that either the height or width will be more or less the same value as the wanted one (that's why I round both values in the function) and the aspect ratio will be the same as the original image

CoreTests/Extensions/UIKit/UIImageSpec.swift
nahuelwolox: Yes, but I don't know if it's useful... The useful cases would be if the image is 4:3, 16:10 or 16:9, but for other random non-standard values the `ratio` wouldn't tell you anything and you would use the size

CoreTests/Extensions/UIKit/UIImageSpec.swift
danielaRiesgo: But if we know both inputs, can't be extrapolate the output?
Like: we have a 50x50 and ask for a 250x346. Since it's the biggest size that fits that wanted one maintaining aspect ratio, it will use the whole width but not height. and so it will be 250x250, right?

CoreTests/Extensions/UIKit/UIImageSpec.swift
danielaRiesgo: I don't know, maybe you want to constraint the image view to the image original ratio. although the image view gives you many ways of handling the image without needing that. Up to you.

nahuelwolox:
Update: I added a parameter to select if you want to use the screen scale for resizing or resize to actual pixels. This can't be tested automatically because the internal representation of the size of an image is in points and not in real pixels, but I tested in other project creating several `.png` files and it works as expected: If you use the screen scale the resulting size of the image would be dependant on the device you're using, if you choose to use real pixels then the resize image will have the pixel per pixel size you specified.

Core/Extensions/UIKit/UIImage.swift
```diff
@@ -15,14 +15,20 @@ public extension UIImage {
...
+        Set to true if you want to make the size relative to the scale of the screen 
+            (So the resized image looks the same relative size on all devices, you probably want this if you're showing the resized image).
+        Set to false if you want the size to be actual pixels 
+            (This is useful to generate a resized image of a particular size in real pixels to send to API regardless of the device screen scale).
      - parameter cornerRadius: The cornerRadius to be used. Any number lower or equal to zero will not add cornerRadius. Default: 0.0
      
      */
     
-    public func resizedTo(_ wantedSize: CGSize, cornerRadius: CGFloat = 0.0) -> UIImage {
+    public func resizedTo(_ wantedSize: CGSize, useScreenScale: Bool = true, cornerRadius: CGFloat = 0.0) -> UIImage {
```
danielaRiesgo: I think there is a card to make all bools enum, maybe you can add this function to the card to change later so it is something like `scale: ImageSizeScale = .screen` (or .pixel)

CoreTests/Extensions/UIKit/UIImageSpec.swift
```diff
@@ -73,7 +73,7 @@ public class UIImageSpec: QuickSpec {
                         let wantedSize = CGSize(width: 15, height: 29)
                         let resizedImage = originalImage.resizedTo(wantedSize)
                         expect(self.ratio(of: originalImage)).to(equal(self.ratio(of: resizedImage)))
-                        expect(self.hasMatchingHeightOrWidth(sizeOne: resizedImage.size, sizeTwo: wantedSize)).to(beTrue())
+                        expect(self.hasMatchingWidth(sizeOne: resizedImage.size, sizeTwo: wantedSize)).to(beTrue())
```
danielaRiesgo: Just to test another case I would make here a case were the height matches and not the width.

danielaRiesgo:
@nahuelwolox check, that according to bitrise there is an error in the tests: `fatal error: unexpectedly found nil while unwrapping an Optional value`

Core/Extensions/UIKit/UIImage.swift
```diff
@@ -0,0 +1,65 @@
...
+     
+     */
+    
+    public func resizedTo(_ wantedSize: CGSize, useScreenScale: Bool = true, cornerRadius: CGFloat = 0.0) -> UIImage {
+        let newSize = size.resizedMaintainingRatio(wantedSize: wantedSize)
+        let scale: CGFloat = useScreenScale ? 0.0 : 1.0
+        
+        UIGraphicsBeginImageContextWithOptions(newSize, false, scale)
+        if cornerRadius > 0 {
+            let path = UIBezierPath(
```
pablog: I'd do:

```
let roundedRect = CGRect(origin: CGPoint.zero, size: newSize)
let path = UIBezierPath(roundedRect: roundedRect, cornerRadius: cornerRadius)
path.addClip()
```

Core/Extensions/UIKit/UIImage.swift
```diff
@@ -0,0 +1,65 @@
...
+                        roundedRect:CGRect(x: 0, y: 0, width: newSize.width, height: newSize.height),
+                        cornerRadius: cornerRadius)
+            path.addClip()
+        }
+        draw(in: CGRect(x: 0, y: 0, width: newSize.width, height: newSize.height))
```
pablog: I'd do the same with this `CGRect`, you may reuse the one declared before if you declare it outside the `if` scope.

Core/Extensions/UIKit/UIImage.swift
```diff
@@ -0,0 +1,65 @@
...
+        let newImage: UIImage = UIGraphicsGetImageFromCurrentImageContext()!
```
pablog: Don't need to specify the type.

CoreTests/Extensions/UIKit/UIImageSpec.swift
pablog: As @danielaRiesgo says, I'd move this to `UIImage.swift`: `public var aspectRatio: CGFloat`

Core/Extensions/UIKit/UIImage.swift
```diff
@@ -0,0 +1,65 @@
...
+     - parameter useScreenScale: Determines if the screen scale will be used when creating the image context. Default: true.
+        Set to true if you want to make the size relative to the scale of the screen 
+            (So the resized image looks the same relative size on all devices, you probably want this if you're showing the resized image).
+        Set to false if you want the size to be actual pixels 
+            (This is useful to generate a resized image of a particular size in real pixels to send to API regardless of the device screen scale).
+     - parameter cornerRadius: The cornerRadius to be used. Any number lower or equal to zero will not add cornerRadius. Default: 0.0
+     
+     */
+    
+    public func resizedTo(_ wantedSize: CGSize, useScreenScale: Bool = true, cornerRadius: CGFloat = 0.0) -> UIImage {
```
pablog: I'm not understanding the parameter `useScreenScale`. It can conflict with the first parameter. What happens if they are not compatible? I'd remove the parameter `useScreenScale` from here. If the user wants the scaling to be proportional to the screen so he should take care of it and provide this function the correct size. I think this extension should not be responsible of this.

pablog: I'd add a new parameter here: `capInsets: UIEdgeInsets`.

I have this function in my project:

```
func resize(withCapInsets capInsets: UIEdgeInsets, toSize newSize: CGSize) -> UIImage {
  UIGraphicsBeginImageContextWithOptions(newSize, false, 0.0)
  resizableImage(withCapInsets: capInsets).draw(in: CGRect(origin: .zero, size: newSize))
  let newImage = UIGraphicsGetImageFromCurrentImageContext()!
  UIGraphicsEndImageContext()
  return newImage
}
```

Can you add this parameter and edit the line?

`draw....` to ` resizableImage(withCapInsets: capInsets).draw...`

Take a look at [this link](https://www.natashatherobot.com/ios-stretchable-button-uiedgeinsetsmake/) to know what this is useful for.

nahuelwolox:
@danielaRiesgo @FranDepascuali we were talking with @pablog about exposing 2 methods, one that takes a size and doesn't respect the aspect ratio, and one that takes a multiplier and respects the aspect ratio. The only problem with that is the case when you want to fit an image in a container respecting the aspect ratio (for example if you have a fixed size container and you want to fit an image that API sends you that can have any size). In that case the user will have to calculate the multiplier to use based on both sizes. 
That could be resolved by adding another boolean in the method that takes a size to select if you want to respect the aspect ratio or not. 
What do you think?

danielaRiesgo:
I like adding a function that doesn't respect the aspect ratio. About the one with the multiplier, I agree that it isn't enough for the case where you want to fit a container and calculating the multiplier for that case is the part of the trouble that we are trying to facilitate. So I believe just one function with the boolean is enough (if the user wants a multiplier then he will just multiply he size which is easy)

Core/Extensions/UIKit/UIImage.swift
```diff
@@ -11,36 +11,64 @@ import UIKit
...
     
     /**
-     Returns a resized copy of the image maintaining it's aspect ratio, with cornerRadius if used.
+     Returns a resized copy of the image, with cornerRadius if used.
      
-     - parameter wantedSize: The wanted size to fit the new image in.
-        Note that the size of the resulting image will be the closest size to the one specified that maintains the aspect ratio of the original image.
+     - parameter toSize: The wanted size to fit the new image in.
+     - parameter maintainAspectRatio: Determines if the resulting image will keep the aspect ratio of the original image, 
+            If set to true the size of the resulting image will be the closest size to the one specified that maintains the aspect ratio of the original image.
```
danielaRiesgo: `closest smaller size`

Core/Extensions/UIKit/UIImage.swift
```diff
@@ -11,36 +11,64 @@ import UIKit
...
         if cornerRadius > 0 {
-            let path = UIBezierPath(
-                        roundedRect:CGRect(x: 0, y: 0, width: newSize.width, height: newSize.height),
-                        cornerRadius: cornerRadius)
+            let path = UIBezierPath(roundedRect:imageRect, cornerRadius: cornerRadius)
             path.addClip()
         }
-        draw(in: CGRect(x: 0, y: 0, width: newSize.width, height: newSize.height))
-        let newImage: UIImage = UIGraphicsGetImageFromCurrentImageContext()!
+        resizableImage(withCapInsets: capInsets).draw(in: imageRect)
```
danielaRiesgo: Where does this `capInsets` come from?

Core/Extensions/UIKit/UIImage.swift
```diff
@@ -11,36 +11,64 @@ import UIKit
...
+        let newImage = UIGraphicsGetImageFromCurrentImageContext()!
         UIGraphicsEndImageContext()
         
         return newImage
     }
+    
+    /**
+     Returns a resized copy of the image. In case that the new size is smaller it will crop it, 
+        in case that it is bigger it will replicate it to fill the desired size, ignoring the points specified by capInsets.
```
danielaRiesgo: What do you mean by ignoring ?
Why not add cap insets to the above function? What if the user wants an image of certain size maximum but maintaining aspect ratio that has certain insets? Could that happen?

Core/Extensions/UIKit/UIImage.swift
```diff
@@ -57,7 +85,7 @@ fileprivate extension CGSize {
     fileprivate func resizedMaintainingRatio(wantedSize: CGSize) -> CGSize {
         let widthFactor = width / wantedSize.width
         let heightFactor = height / wantedSize.height
-        let resizeFactor = height > width ? heightFactor : widthFactor
+        let resizeFactor = wantedSize.height < wantedSize.width ? heightFactor : widthFactor
```
danielaRiesgo: Why this?
If the image is 50x150 and the wanted size is 150x50, it should stretch the height and then accordingly stretch the width, right? and the height is greater in the original image, not the wanted size.

nahuelwolox: In that example the resizeFactor would be the widthFactor = 50/150 which is ok because I want to fit the original image in the new container so it would have to fit the width and scale the height proportionally

Core/Extensions/UIKit/UIImage.swift
nahuelwolox: I'm not sure if that's a case that could happen... Resize with cap insets works in a different way that a normal resize would: It will crop the image if the wanted size is smaller, and it will replicate the image if the wanted size is bigger than the original, you could say it resizes the canvas and fills it. The Cap insets set a number of points of the image you don't want to move, replicate or crop regardless of the wanted size. This is mainly used for 9-slicing an image that works as a menu or button background. Check this example: https://www.natashatherobot.com/ios-stretchable-button-uiedgeinsetsmake/

Core/Extensions/UIKit/UIImage.swift
nahuelwolox: I'm not sure I understand the question, the user sets the cap insets

Core/Extensions/UIKit/UIImage.swift
nahuelwolox: I'll change this so the user can select between aspectFit and aspectFill modes

Core/Extensions/UIKit/UIImage.swift
nahuelwolox: AH! I finally understood what you meant, that's unintended but somehow it compiles. I'll remove it.

Core/Extensions/UIKit/UIImage.swift
```diff
@@ -36,7 +36,7 @@ public extension UIImage {
             let path = UIBezierPath(roundedRect:imageRect, cornerRadius: cornerRadius)
```
danielaRiesgo: Add space before `imageRect `.

danielaRiesgo:
@nahuelwolox what about this?

CoreTests/Extensions/UIKit/UIImageSpec.swift
```diff
@@ -0,0 +1,113 @@
...
+    
+    override public func spec() {
+        
+        var originalImage: UIImage!
+        
+        describe("#resized") {
+            
+            beforeEach {
+                let bundle = Bundle(for: type(of: self))
+                originalImage = UIImage(named: "50by50", in: bundle, compatibleWith: nil)
```
danielaRiesgo: Can't you use .none  here?

woloxci:

:white_check_mark: Code Coverage (Slather): 87.0
:white_check_mark: Code Quality (Swiftlint): 90.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/image-extension) for more details.

woloxci:

:white_check_mark: Code Coverage (Slather): 86.27
:white_check_mark: Code Quality (Swiftlint): 90.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/image-extension) for more details.

Issue #73, opened 2017-06-13
Add support for Swift 3.2


cristianames92:
Support for Xcode 9 starting at least with Swift 3.2 would be nice.

damfinkel:
@cristianames92 I think we want to migrate directly to Swift 4, right? Make a new issue and close this one.

PR #74, opened 2017-06-13
Swift 3.2 compatibility

cristianames92:
## Issue ##
https://github.com/Wolox/wolmo-core-ios/issues/73

This brings support for Swift 3.2 using Xcode 9. Next step would be migrate to Swift 4

## Known issues ##
The ReactiveCocoa reference in the Cartfile is directly pointing to an alpha release, something that should be changed to the `6.0` series when they release a final version of the library.


CoreTests/Extensions/Foundation/BundleSpec.swift
```diff
@@ -268,7 +268,7 @@ public class BundleSpec: QuickSpec {
                     
                     it("should return .none") {
                         let value = bundle.getInt(from: "My Personal Boolean Key")
-                        expect(value).to(beNil())
+                        expect(value).to(equal(1))
```
danielaRiesgo: Why should this change?

script/.env
```diff
@@ -1,6 +1,6 @@
-REQUIRED_XCODE_VERSION=8.3
+REQUIRED_XCODE_VERSION=9.0
 REQUIRED_SWIFTLINT_VERSION=0.16.1
```
danielaRiesgo: Isn't there an updated swiftlint version for swift 2.3 additions/changes?

pablog:
@cristianames92

CoreTests/Extensions/Foundation/BundleSpec.swift
cristianames92: It seems that the new changes involved this new interpretation of the boolean value. If you try to cast it to an int it will apparently return 0 or 1 according the bool value.

script/.env
cristianames92: Not yet, but Realm has more recent releases for the old swift 3.0 versions. They are now under the version 0.19.0. I didn't change it because there is no new version for Swift 3.2.

CoreTests/Extensions/Foundation/BundleSpec.swift
danielaRiesgo: Okay, so then we should update the documentation and test description.

script/.env
danielaRiesgo: Okay. I would update it to 0.19.0 none the less.

CoreTests/Extensions/Foundation/BundleSpec.swift
```diff
@@ -264,9 +264,9 @@ public class BundleSpec: QuickSpec {
                     
                 }
                 
-                context("when the value is not the expected type") {
+                context("when the value can be converted in other types") {
```
danielaRiesgo: Shouldn't we then add another test for the not expected type failing to None?

cristianames92:
The project will still failing in Bitrise due to the version of XCode in the server

pablog:
@damfinkel could you pelase check if bitrise is failing because of the same you found a workaround?

danielaRiesgo:
I can't see the logs, they never end loading. @damfinkel any clue?

damfinkel:
Closing this pull request, will migrate directly to swift 4

PR #80, opened 2017-09-11
added an extension to use string literals to create URLs

damfinkel:
## Summary ##

Added an extension to support string literals for URLs


pablog:
Remember to reference the issue in the description so it closes when the PR is merged: https://github.com/blog/1506-closing-issues-via-pull-requests

Core/Extensions/Foundation/URL.swift
```diff
@@ -34,3 +34,19 @@ public extension URL {
...
+    public init(unicodeScalarLiteral value: String) {
+        self.init(stringLiteral: value)
+    }
+    
+    public init(extendedGraphemeClusterLiteral value: String) {
+        self.init(stringLiteral: value)
+    }
+    
+    public init(stringLiteral value: StringLiteralType) {
+        self = URL(string: "\(value)")!
```
danielaRiesgo: Couldn't this fail? I think the twit where we took this from had something so that we had an error message or something

CoreTests/Extensions/Foundation/URLSpec.swift
```diff
@@ -74,5 +74,18 @@ public class URLSpec: QuickSpec {
                 }
             }
         }
+        
+        describe("#stringLiteral") {
```
danielaRiesgo: And the other two are used at what cases?

damfinkel: The other two are needed to comply to the ExpressibleByStringLiteral protocol.

Core/Extensions/Foundation/URL.swift
damfinkel: That code doesn't work. URL doesn't have a "require" method. Besides, I think it's best if it fails when the url isn't a valid one. It probably means that the content the app should be showing is not valid

danielaRiesgo: Sure, it should fail because it won't work. But maybe with a better message that "found nil".

woloxci:

:white_check_mark: Code Coverage (Slather): 87.37


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/url-string-literal) for more details.

PR #83, opened 2017-10-12
View gradients

danielaRiesgo:
## Summary ##

Adding a wrapper to easily handle gradients in views.

Tested in Fitivity and working.

### DIFF
--- a/Core/Extensions/UIKit/UIView/Gradient.swift
+++ b/Core/Extensions/UIKit/UIView/Gradient.swift
@@ -29,12 +29,12 @@ private extension GradientDirection {
         switch self {
         case .leftToRight: return CGPoint(x: 0, y: 0.5)
         case .rightToLeft: return CGPoint(x: 1, y: 0.5)
-        case .topToBottom: return CGPoint(x: 0.5, y: 1)
-        case .bottomToTop: return CGPoint(x: 0.5, y: 0)
-        case .topLeftToBottomRight: return CGPoint(x: 0, y: 1)
-        case .bottomRightToTopLeft: return CGPoint(x: 1, y: 0)
-        case .topRightToBottomLeft: return CGPoint(x: 1, y: 1)
-        case .bottomLeftToTopRight: return CGPoint(x: 0, y: 0)
+        case .topToBottom: return CGPoint(x: 0.5, y: 0)
+        case .bottomToTop: return CGPoint(x: 0.5, y: 1)
+        case .topLeftToBottomRight: return CGPoint(x: 0, y: 0)
+        case .bottomRightToTopLeft: return CGPoint(x: 1, y: 1)
+        case .topRightToBottomLeft: return CGPoint(x: 1, y: 0)
+        case .bottomLeftToTopRight: return CGPoint(x: 0, y: 1)
         }
     }
     
@@ -42,12 +42,12 @@ private extension GradientDirection {
         switch self {
         case .leftToRight: return CGPoint(x: 1, y: 0.5)
         case .rightToLeft: return CGPoint(x: 0, y: 0.5)
-        case .topToBottom: return CGPoint(x: 0.5, y: 0)
-        case .bottomToTop: return CGPoint(x: 0.5, y: 1)
-        case .topLeftToBottomRight: return CGPoint(x: 1, y: 0)
-        case .bottomRightToTopLeft: return CGPoint(x: 0, y: 1)
-        case .topRightToBottomLeft: return CGPoint(x: 0, y: 0)
-        case .bottomLeftToTopRight: return CGPoint(x: 1, y: 1)
+        case .topToBottom: return CGPoint(x: 0.5, y: 1)
+        case .bottomToTop: return CGPoint(x: 0.5, y: 0)
+        case .topLeftToBottomRight: return CGPoint(x: 1, y: 1)
+        case .bottomRightToTopLeft: return CGPoint(x: 0, y: 0)
+        case .topRightToBottomLeft: return CGPoint(x: 0, y: 1)
+        case .bottomLeftToTopRight: return CGPoint(x: 1, y: 0)
         }
     }
     
@@ -59,7 +59,7 @@ private extension GradientDirection {
  */
 public struct ViewGradient {
     
-    private let layer: CAGradientLayer
+    fileprivate let layer: CAGradientLayer
     
     /**
      Initializes an inmutable ViewGradient.
--- a/Core/Extensions/UIKit/UIView/Gradient.swift
+++ b/Core/Extensions/UIKit/UIView/Gradient.swift
@@ -53,73 +53,98 @@ fileprivate extension GradientDirection {
 }
 
 /**
+     Represents a color that takes place ina gradient that can be applied to a view.
+     It has a UIColor and a location which indicates where that color should be placed in the gradient.
+ */
+public struct GradientColor {
+
+    // Color to use in gradient.
+    public let color: UIColor
+    // Location where to place color inside the gradient.
+    // Number between 0 and 1 (inclusive).
+    public let location: Float
+
+    /**
+     Initializes an inmutable GradientColor.
+
+     - parameter color: UIColor to use.
+     - parameter location: Locations of the center of the color through out the gradient.
+     A location is any possible number between 0 and 1, being 0 the start of the gradient
+     and 1 the end of the gradient in the direction specified.
+     - warning: If the location isn't between 0 and 1 (inclusive) the init will fail.
+     */
+    public init?(color: UIColor, location: Float) {
+        if location < 0 || location > 1 {
+            return nil
+        }
+        self.color = color
+        self.location = location
+    }
+
+}
+
+/**
     Represents a gradient that can be applied to a view.
     It can have many colors distributed in many ways and directions.
-    There can only be one gradient at a time in a view.
+
+    - note: There can only be one gradient at a time in a view.
  */
 public struct ViewGradient {
     
     // Direction of the gradient.
     public let direction: GradientDirection
-    // Colors involved in the order involved.
-    public let colors: [UIColor]
-    // Locations where the colors are placed.
-    public var locations: [Float] {
-        return layer.locations!.map { $0.floatValue }
-    }
+    // GradientColors involved in the order involved.
+    public let colors: [GradientColor]
     
     fileprivate let layer: CAGradientLayer
-    
+
+    /**
+     Initializes an inmutable ViewGradient.
+
+     - parameter colors: Array of GradientColors to be used.
+     - parameter direction: Direction in which the gradient should develop.
+     */
+    public init(colors: [GradientColor], direction: GradientDirection) {
+        self.colors = colors.sorted { g1, g2 in g1.location < g2.location }
+        self.direction = direction
+
+        layer = CAGradientLayer()
+        layer.anchorPoint = .zero
+
+        layer.colors = self.colors.map { $0.color.cgColor }
+        layer.locations = self.colors.map { NSNumber(value: $0.location) }
+
+        layer.startPoint = direction.startPoint
+        layer.endPoint = direction.endPoint
+    }
+
     /**
      Initializes an inmutable ViewGradient.
      
      - parameter colors: Array of UIColors in the order in which to place them.
-        There should be more than one.
+        They will be evenly separated.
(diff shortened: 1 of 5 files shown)


Core/Extensions/UIKit/UIView/Gradient.swift
```diff
@@ -0,0 +1,135 @@
...
+        gradientLayer.bounds = bounds
+        reactive.producer(forKeyPath: "bounds")
+            .take(duringLifetimeOf: self)
+            .take(duringLifetimeOf: gradientLayer)
+            .map { $0 as! CGRect }
+            .on(value: { [unowned gradientLayer] in gradientLayer.bounds = $0})
+            .start()
+        layer.insertSublayer(gradientLayer, at: 0)
+    }
+
```
pablog: Remove some of these empty lines

pablog: @danielaRiesgo I mean the empty lines between the function and the end of the extension

danielaRiesgo: Yes, done now.

Core/Extensions/UIKit/UIView/Gradient.swift
```diff
@@ -0,0 +1,132 @@
...
+    case bottomLeftToTopRight
+}
+
+private extension GradientDirection {
+    
+    var startPoint: CGPoint {
+        switch self {
+        case .leftToRight: return CGPoint(x: 0, y: 0.5)
+        case .rightToLeft: return CGPoint(x: 1, y: 0.5)
+        case .topToBottom: return CGPoint(x: 0.5, y: 1)
```
damfinkel: Are you sure about the vertical ones? I'm guessing the GradientDirection describes the direction from first color to last color. Given that topY < bottomY, check if the startingPoint of .topToBottom should be CGPoint(x:0.5, y: 0). Same with the rest of vertical gradients.

Core/Extensions/UIKit/UIView/Gradient.swift
```diff
@@ -0,0 +1,132 @@
...
+    public func apply(gradient: ViewGradient) {
+        let gradientLayer = gradient.layer
+        gradientLayer.bounds = bounds
+        reactive.producer(forKeyPath: "bounds")
+            .take(duringLifetimeOf: self)
+            .take(duringLifetimeOf: gradientLayer)
+            .map { $0 as! CGRect }
+            .on(value: { [unowned gradientLayer] in gradientLayer.bounds = $0})
+            .start()
+        layer.insertSublayer(gradientLayer, at: 0)
```
damfinkel: In Kanvas I had issues with gradients when rotating the screen in some cases (depending if the views with the gradients rotated or not). There should be a way to replace and/or remove the gradientView.

Core/Extensions/UIKit/UIView/Gradient.swift
danielaRiesgo: Yeah ! You are right. I changed it in Fitivity and forgot it here.

Core/Extensions/UIKit/UIView/Gradient.swift
danielaRiesgo: I tried rotating the screen and the gradient rotated too. But If you set that the app is only portrait what happens.
If you don't want the gradient to change rotation with the view, then yes, I guess you should add another gradient.
The only question with how to remove it, is that you have to store something.
For example, or the user must keep the gradient and you search in the view for the layer that matches and remove it.
Or the gradient should be created with a view and store the view to be applied on or removed of, but the user would still need to save the GradientView.

I guess I go with the first one. And if you want to change it, you remove this view gradient and create another one and apply it.
What do you say?

woloxci:

:x: Code Coverage (Slather): 77.88 - Minimum: 85.0
:x: Code Quality (Swiftlint): 88.0 - Minimum: 90.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/view-gradients) for more details.

Core/Extensions/UIKit/UIView/Gradient.swift
danielaRiesgo: I tried setting only portrait for example, and as the view doesn't rotate, the gradient doesn't either. If you want to see how it looks, just let me know.

woloxci:

:white_check_mark: Code Coverage (Slather): 86.27
:x: Code Quality (Swiftlint): 88.0 - Minimum: 90.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/view-gradients) for more details.

Core/Extensions/UIKit/UIView/Gradient.swift
damfinkel: When the view rotates you might want to change the gradient rotation or remove it completely, I still think we need at least a remove. In Kanvas the issue was that a certain view didn't rotate with the same angle as the phone, so the gradient stayed the same when it actually had to change. 

A way you can remove a layer without storing anything is using its name property. It may not be the cleanest though: 
https://stackoverflow.com/a/46144154/7915127

woloxci:

:white_check_mark: Code Coverage (Slather): 86.27
:x: Code Quality (Swiftlint): 86.0 - Minimum: 90.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/view-gradients) for more details.

damfinkel:
@danielaRiesgo just curious, why did you went for the associated object approach at the end?

Core/Extensions/UIKit/UIView/Gradient.swift
```diff
@@ -0,0 +1,174 @@
...
+     - parameter direction: Direction in which the gradient should develop.
+     - parameter locations: Locations of the center of each color through out the screen.
+        A location is any possible number between 0 and 1, being 0 the start of the gradient
+        and 1 the end of the gradient in the direction specified.
+        By default, it spreads the colors uniformly.
+        If passed a value, it should have as many locations as it does colors.
+    */
+    public init(colors: [UIColor], direction: GradientDirection, locations: [Float]? = .none) {
+        precondition(colors.count > 1)
+        if let locs = locations {
```
pablog: In this cases you can use `if let locations = locations`. This allows you to unwrap the property and keep using a consistent name from then on.

Core/Extensions/UIKit/UIView/Gradient.swift
```diff
@@ -0,0 +1,174 @@
...
+        }
+        
+        self.colors = colors
+        self.direction = direction
+        
+        layer = CAGradientLayer()
+        layer.anchorPoint = .zero
+        layer.colors = colors.map { $0.cgColor }
+        
+        setUpLocations(locations: locations, numberOfColors: colors.count)
```
pablog: I'd make `setUpLocations` return `[NSNumber]`, so you assign everything in this function. I think is more clear.

Also, I'd change the name `setUpLocations` to:

`func locations(from locations: [Float]?, numberOfColors: Int) -> [NSNumber]` (or something like this).

Core/Extensions/UIKit/UIView/Gradient.swift
```diff
@@ -0,0 +1,174 @@
...
+            let colorProportion = 1.0 / Double(numberOfColors - 1)
+            for i in 0...numberOfColors {
+                let location = Double(i) * colorProportion
+                locs.append(NSNumber(value: location))
+            }
+        }
+        layer.locations = locs
+    }
+    
+    private func setUpDirection(_ direction: GradientDirection) {
```
pablog: I don't mind this, but following the line of assigning everything in `init`, you may consider not having this function. But as you prefer.

danielaRiesgo:
@damfinkel I wanted to avoid something that could fail, like giving the layer a name, that maybe the user can use the same name (although it would be weird) or that if the user wanted to do strange things to hack the framework he could by using the name.
Moreover, using a name would mean to search through all the layers the view has to be able to find it. Although I think a view may have much more layers than we manually add, it would not be such a problem since you would only search for it for deleting most probably, but if the user tries to read the gradient property many times for any reason, it would not be performant.

woloxci:

:white_check_mark: Code Coverage (Slather): 86.27
:white_check_mark: Code Quality (Swiftlint): 90.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/view-gradients) for more details.

woloxci:

:white_check_mark: Code Coverage (Slather): 86.27
:white_check_mark: Code Quality (Swiftlint): 90.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/view-gradients) for more details.

woloxci:

:white_check_mark: Code Coverage (Slather): 86.27
:white_check_mark: Code Quality (Swiftlint): 90.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/view-gradients) for more details.

danielaRiesgo:
@damfinkel any other comment?

Core/Extensions/UIKit/UIView/Gradient.swift
```diff
@@ -0,0 +1,172 @@
...
+    It can have many colors distributed in many ways and directions.
+    There can only be one gradient at a time in a view.
+ */
+public struct ViewGradient {
+    
+    // Direction of the gradient.
+    public let direction: GradientDirection
+    // Colors involved in the order involved.
+    public let colors: [UIColor]
+    // Locations where the colors are placed.
```
cristianames92: @danielaRiesgo  You can add `from 0 to 1` in this comment

Core/Extensions/UIKit/UIView/Gradient.swift
```diff
@@ -0,0 +1,172 @@
...
+     - parameter colors: Array of UIColors in the order in which to place them.
+        There should be more than one.
+     - parameter direction: Direction in which the gradient should develop.
+     - parameter locations: Locations of the center of each color through out the screen.
+        A location is any possible number between 0 and 1, being 0 the start of the gradient
+        and 1 the end of the gradient in the direction specified.
+        By default, it spreads the colors uniformly.
+        If passed a value, it should have as many locations as it does colors.
+    */
+    public init(colors: [UIColor], direction: GradientDirection, locations: [Float]? = .none) {
```
cristianames92: I am not very fan of the precondition rather than forcing a type check to ensure right behavior at building time. In this particular case I would use something like this:
```
public init(direction: GradientDirection, initialColor: UIColor, finalColor: UIColor, middleColors: [GradientColor]) {
...
}

class GradientColor {
    let color: UIColor
    let location: Float
}
```

This should remove all the preconditions but the one where you check the Float value. I would see the way to avoid checking this at runtime too, but right now I don't have any workaround on this

danielaRiesgo: I agree about avoiding runtime errors. And I like the GradientColor as a struct, but the disadvantage is that then locations can't be optionals. But we can create two public inits if not.
What I don't particularly like is the initial final and middle. I don't think it's friendly at all for the user since he will most surely have them all in one array, and he will have to do: `initial: colors.first!, last: colors.last!, middles: colors[1..<colors.count]`.
What we can do is directly avoid de precondition of count being 2 or more. Maybe being 1 and add a clear color by default, and if it's 0 then everything clear color. Which doesn't make any sense, but it doesn't make any sense for the user to use it with 0 colors so it would be his fault, we are giving him what he asked.

woloxci:

:white_check_mark: Code Coverage (Slather): 86.27
:white_check_mark: Code Quality (Swiftlint): 90.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/view-gradients) for more details.

woloxci:

:white_check_mark: Code Coverage (Slather): 86.69
:white_check_mark: Code Quality (Swiftlint): 90.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/view-gradients) for more details.

woloxci:

:white_check_mark: Code Coverage (Slather): 86.69
:white_check_mark: Code Quality (Swiftlint): 90.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/view-gradients) for more details.

PR #93, opened 2018-04-03
Added animations extensions

aducret:
## Summary ##
This PR adds animations extensions and a demo project.

## Known issues ##
@danielaRiesgo comments:
- Ideally you could put a segmented controller with one tab for each animation so the user can change to see it without recompiling.
- I think there could be a way of reusing code through different types of animation but I couldn't think of something so clearly.

We should keep working on this improvements.


danielaRiesgo:
@aducret just take care of ignoring files that cannot be tested in the `.slather.yml` file and correct any style for swiftlint to pass.

aducret:
@danielaRiesgo Regarding to the protocols and functions exposed to the user I prefer to wait until we fix the issues that I added because I think the code will be change. 

And about using the old demo project it could be a nice idea, I can add another issue to do this.
Maybe we can add a table and the user can choose the demo which they want.

aducret:
@danielaRiesgo Updated `.slather.yml` and swiftlint warnings! Check if is it ok pls 😀

woloxci:

:white_check_mark: Code Coverage (Slather): 86.0
:white_check_mark: Code Quality (Swiftlint): 100.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/animations) for more details.

danielaRiesgo:
You can now merge it ! :D 🎉

Issue #60, opened 2017-05-22
Making Booleans be Enums


danielaRiesgo:
In UIButton.swift, `setTitle` function, instead of having imageOnRight it should receive a direction which could be either left or right.
In UIView.swift, in all functions with `useConstraints` boolean, it should be turned to an enum that's either frame or constraint. Maybe it could be called `LayoutMode`.

PR #94, opened 2018-04-12
Replaced useConstraints boolean to LayoutMode enum

florenciarosental:
## Summary ##
Trello card: None
Issue: https://github.com/Wolox/wolmo-core-ios/issues/60

Changed all functions with useConstraints boolean in UIView.swift to an enum called LayoutMode.
Replaced the each if that handled those booleans to a switch.


Core/Extensions/UIKit/UIView/UIView.swift
```diff
@@ -40,6 +40,15 @@ public enum BorderPosition {
     }
 }
 
+/**
+ Indicates if a view will use constrains or frame.
```
danielaRiesgo: I would say `Indicates the way in which the layout is set`, because maybe now is only constraints or frames but it may change (although I don't think so). And in any case, you can see the options inside the enum directly.

Core/Extensions/UIKit/UIView/UIView.swift
```diff
@@ -101,22 +110,24 @@ public extension UIView {
             By default, 0.
      - parameter rightOffset: Offset from the view's right border to where the border should start.
             By default, 0.
-     - parameter useConstraints: Boolean indicating whether to use constraints or frames. By default, true.
+     - parameter layout: Enum indicating whether to use constraints or frames. By default, constraints.
```
danielaRiesgo: In the same way: `indicating the layout mode. By default, .constraints`

Core/Extensions/UIKit/UIView/UIView.swift
```diff
@@ -101,22 +110,24 @@ public extension UIView {
...
      
      - note: If you decide to use constraints to determine the size, self's frame doesn't need to be final.
              Because of this, it can be used in `loadView()`, `viewDidLoad()` or `viewWillAppear(animated:)`.
              We strongly recommend to work with constraints as a better practice than frames.
      */
     public func add(top border: BorderViewProperties,
                     withLeftOffset left: CGFloat = 0, rightOffset right: CGFloat = 0,
-                    useConstraints: Bool = true) -> BorderView {
+                    layout: LayoutMode = LayoutMode.constraints) -> BorderView {
```
danielaRiesgo: You can just write: `layout: LayoutMode = .constraints`

script/.env
```diff
@@ -1,4 +1,4 @@
-REQUIRED_XCODE_VERSION=9.0
+REQUIRED_XCODE_VERSION=9.2
```
danielaRiesgo: Maybe you could add an issue in `guidomb/ios-scripts` so that the scripts use this xcode version as a minimum instead of a fixed version. And even better make a PR for that change :P

florenciarosental: Great, I will add it.

danielaRiesgo:
Congrats on your first contribution ! 🎉

woloxci:

:white_check_mark: Code Coverage (Slather): 86.0


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/replacement-of-boolean-to-enum) for more details.

PR #120, opened 2019-01-03
Screen edge pan gesture with closure

gmazzei:
## Summary ##

An extension for `UIView` that allows adding a `UIScreenEdgePanGestureRecognizer` with a closure (instead of using a selector).

#### Example
```Swift
myView.addScreenEdgePanGestureRecognizer { recognizer in
    print("View pinched!")
}
```

```Swift
// Works only on left edge
myView.addScreenEdgePanGestureRecognizer(edges: .left) { recognizer in
    print("View pinched!")
}
```

```Swift
// Works only on left and right edges
myView.addScreenEdgePanGestureRecognizer(edges: [.left, .right]) { recognizer in
    print("View pinched!")
}
```


gmazzei:
Thanks! Although merging is still blocked for me :( You can merge it when you want

PR #123, opened 2019-01-17
Shake Effect

Nick-m96:
## Summary ##

Add shake effect extension to UIViews

## Screencast ##

mov1

## Known issues ##

No known issues

### DIFF
--- a/WolmoCore/Extensions/Animations/UIViewExtension.swift
+++ b/WolmoCore/Extensions/Animations/UIViewExtension.swift
@@ -33,8 +33,8 @@ public extension UIView {
         animation.duration = duration
         animation.repeatCount = 3
         animation.autoreverses = true
-        animation.fromValue = CGPoint(x: self.center.x - 4, y: self.center.y)
-        animation.toValue = CGPoint(x: self.center.x + 4, y: self.center.y)
+        animation.fromValue = CGPoint(x: center.x - 4, y: center.y)
+        animation.toValue = CGPoint(x: center.x + 4, y: center.y)
         
         self.layer.add(animation, forKey: "position")
     }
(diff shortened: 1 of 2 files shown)


gmazzei:
Great, I'll review it soon!

WolmoCore/Extensions/Animations/UIViewExtension.swift
```diff
@@ -28,4 +28,15 @@ public extension UIView {
...
     
+    public func shake(withDuration duration: TimeInterval = 0.05) {
+        let animation = CABasicAnimation(keyPath: "position")
+        animation.duration = duration
+        animation.repeatCount = 3
+        animation.autoreverses = true
+        animation.fromValue = CGPoint(x: self.center.x - 4, y: self.center.y)
+        animation.toValue = CGPoint(x: self.center.x + 4, y: self.center.y)
+        
+        self.layer.add(animation, forKey: "position")
```
gmazzei: Remove unnecessary `self`'s here and above.

WolmoCoreDemo/ViewController.swift
```diff
@@ -28,6 +28,7 @@ final internal class ViewController: UIViewController {
         _view.stringsButton.addTarget(self, action: #selector(buttonAction), for: .touchUpInside)
         
         _view.gestureLabel.addTapGestureRecognizer {
```
gmazzei: In this case, we should add `[weak self]` at the beginning of the block to avoid memory issues.

WolmoCoreDemo/ViewController.swift
```diff
@@ -28,6 +28,7 @@ final internal class ViewController: UIViewController {
...
+            self._view.gestureLabel.shake(withDuration: 0.05)
```
gmazzei: In this case, we should add `[weak self]` at the beginning of the block to avoid memory issues.

WolmoCore/Extensions/Animations/UIViewExtension.swift
Nick-m96: Sorry 😅

WolmoCoreDemo/ViewController.swift
Nick-m96: Done 👍

PR #125, opened 2019-01-23
Fix screenEdgePanRecognizer

Nick-m96:
## Summary ##

An extension for UIView that allows adding a UIScreenEdgePanGestureRecognizer with a closure (instead of using a selector). One edge is always required.

##Example##

```
myView.addScreenEdgePanGestureRecognizer(edges: .left) { recognizer in
    print("Edge panned!")
}
```

### DIFF
--- a/WolmoCore/Extensions/UIKit/UIView/GestureRecognizers/ScreenEdgePanRecognizer.swift
+++ b/WolmoCore/Extensions/UIKit/UIView/GestureRecognizers/ScreenEdgePanRecognizer.swift
@@ -34,14 +34,14 @@ public extension UIView {
     
     /**
      Adds a screen-edge-pan gesture recognizer that executes the closure when panned on the edge
-     - Parameter edges: The edges on which this gesture recognizes, relative to the current interface orientation. Default is all edges
+     - Parameter edges: The edge on which this gesture recognizes, relative to the current interface orientation. Default is left edge
      - Parameter action: The closure that will execute when the edge of the view is panned
      */
-    public func addScreenEdgePanGestureRecognizer(edges: UIRectEdge = .all, action: ((UIScreenEdgePanGestureRecognizer) -> Void)?) {
+    public func addScreenEdgePanGestureRecognizer(edge: UIRectEdge, action: ((UIScreenEdgePanGestureRecognizer) -> Void)?) {
         isUserInteractionEnabled = true
         screenEdgePanGestureRecognizerAction = action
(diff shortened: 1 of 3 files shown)


WolmoCore/Extensions/UIKit/UIView/GestureRecognizers/ScreenEdgePanRecognizer.swift
```diff
@@ -34,14 +34,14 @@ public extension UIView {
     
     /**
      Adds a screen-edge-pan gesture recognizer that executes the closure when panned on the edge
-     - Parameter edges: The edges on which this gesture recognizes, relative to the current interface orientation. Default is all edges
+     - Parameter edges: The edge on which this gesture recognizes, relative to the current interface orientation. Default is left edge
```
gmazzei: Change `Parameter edges:` to `Parameter edge:` in this comment

WolmoCore/Extensions/UIKit/UIView/GestureRecognizers/ScreenEdgePanRecognizer.swift
```diff
@@ -34,14 +34,14 @@ public extension UIView {
...
      - Parameter action: The closure that will execute when the edge of the view is panned
      */
-    public func addScreenEdgePanGestureRecognizer(edges: UIRectEdge = .all, action: ((UIScreenEdgePanGestureRecognizer) -> Void)?) {
+    public func addScreenEdgePanGestureRecognizer(edge: UIRectEdge, action: ((UIScreenEdgePanGestureRecognizer) -> Void)?) {
```
gmazzei: If the default value is `.left` then we should write `edge: UIRectEdge = .left` in the method signature

Nick-m96: Should we have a default value?

WolmoCore/Extensions/UIKit/UIView/GestureRecognizers/ScreenEdgePanRecognizer.swift
Nick-m96: Done :)

WolmoCore/Extensions/UIKit/UIView/GestureRecognizers/ScreenEdgePanRecognizer.swift
gmazzei: No, it is not necessary. But the documentation above says  "Default is left edge". You can leave the parameter as mandatory but in that case remove the last sentence.

PR #130, opened 2019-04-10
Add animation to tableView

NickMano:
## Summary ##
Add the animation animateTable to TableViews

## Screencast ##

mov1

## Known issues ##

No known issues

### DIFF
--- a/AnimationDemo/ViewController.swift
+++ b/AnimationDemo/ViewController.swift
@@ -44,6 +44,10 @@ class ViewController: UIViewController {
         draggableView.backgroundColor = UIColor.red
         draggableView.layer.masksToBounds = true
         draggableView.layer.cornerRadius = 23.5
+        
+        view.addScreenEdgePanGestureRecognizer(edge: .left) { recognizer in
+            self.present(TableViewController(), animated: true, completion: .none)
+        }
     }
 
     override func viewDidAppear(_ animated: Bool) {
--- a/AnimationDemo/ViewController.swift
+++ b/AnimationDemo/ViewController.swift
@@ -45,8 +45,8 @@ class ViewController: UIViewController {
         draggableView.layer.masksToBounds = true
         draggableView.layer.cornerRadius = 23.5
         
-        view.addScreenEdgePanGestureRecognizer(edge: .left) { recognizer in
-            self.present(TableViewController(), animated: true, completion: .none)
+        view.addScreenEdgePanGestureRecognizer(edge: .left) { [weak self] _ in
+            self?.present(TableViewController(), animated: true, completion: .none)
         }
     }
 
--- a/WolmoCore/Extensions/Animations/UIViewExtension.swift
+++ b/WolmoCore/Extensions/Animations/UIViewExtension.swift
@@ -103,13 +103,14 @@ public extension UITableView {
         let tableWidth = self.bounds.size.width
         var delayCounter = delay!
         
-        for cell in cells { cell.transform = CGAffineTransform(translationX: tableWidth * CGFloat(directionX!), y: tableHeight * CGFloat(directionY!)) }
+        for cell in cells {
+            cell.transform = CGAffineTransform(translationX: tableWidth * CGFloat(directionX!), y: tableHeight * CGFloat(directionY!))
+        }
 
         for cell in cells {
             UIView.animate(withDuration: duration, delay: delayCounter * delayMultiplier!, usingSpringWithDamping: 0.8, initialSpringVelocity: 0, options: .curveEaseInOut, animations: {
                 cell.transform = CGAffineTransform.identity
             }, completion: nil)
-            print(delayCounter * delayMultiplier!)
             delayCounter += 1
         }
     }
--- a/WolmoCore/Extensions/Animations/UIViewExtension.swift
+++ b/WolmoCore/Extensions/Animations/UIViewExtension.swift
@@ -82,3 +82,36 @@ public extension UIView {
     }
     
 }
+
+public extension UITableView {
+    
+    /**
+     Adds a custom animation to the table that executes the closure when long pressed
+     
+     - Parameter duration: Time in seconds the animation will be executed for cell. Default is 1.75
+     - Parameter delay: Time in seconds the animation will be stopped between cells. Default is 0
+     - Parameter delayMultiplier: Number that the delay will be multiplier. Default is 0.01
+     - Parameter directionX: Direction in axis X that the cell come from the original positio. Default is 1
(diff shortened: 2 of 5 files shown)


woloxci:

:white_check_mark: Code Coverage (Slather): 84.3


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/animation-tableview) for more details.

AnimationDemo/ViewController.swift
```diff
@@ -44,6 +44,10 @@ class ViewController: UIViewController {
         draggableView.backgroundColor = UIColor.red
         draggableView.layer.masksToBounds = true
         draggableView.layer.cornerRadius = 23.5
+        
+        view.addScreenEdgePanGestureRecognizer(edge: .left) { recognizer in
```
gmazzei: If `recognizer` is not used inside the block, you can replace it with an underscore. Also, you can add `[unowned self]` or `[weak self]` here.

WolmoCore/Extensions/Animations/UIViewExtension.swift
```diff
@@ -82,3 +82,36 @@ public extension UIView {
...
+    
+    func animateTable(duration: TimeInterval = 1.75, delay: Double? = 0, delayMultiplier: Double? = 0.01, directionX: Float? = 1, directionY: Float? = 0) {
+        self.reloadData()
+        
+        let cells = self.visibleCells
+        let tableHeight = self.bounds.size.height
+        let tableWidth = self.bounds.size.width
+        var delayCounter = delay!
+        
+        for cell in cells { cell.transform = CGAffineTransform(translationX: tableWidth * CGFloat(directionX!), y: tableHeight * CGFloat(directionY!)) }
```
gmazzei: Can we break this into three lines?
```Swift
for cell in cells { 
    cell.transform = CGAffineTransform(translationX: tableWidth * CGFloat(directionX!), y: tableHeight * CGFloat(directionY!)) 
}
```

WolmoCore/Extensions/Animations/UIViewExtension.swift
```diff
@@ -82,3 +82,36 @@ public extension UIView {
...
+
+        for cell in cells {
+            UIView.animate(withDuration: duration, delay: delayCounter * delayMultiplier!, usingSpringWithDamping: 0.8, initialSpringVelocity: 0, options: .curveEaseInOut, animations: {
+                cell.transform = CGAffineTransform.identity
+            }, completion: nil)
+            print(delayCounter * delayMultiplier!)
```
gmazzei: Delete this `print` if it isn't necessary.

Nick-m96: Sorry for this, it was a test 😅

WolmoCore/Extensions/Animations/UIViewExtension.swift
```diff
@@ -82,3 +82,36 @@ public extension UIView {
...
+        
+        let cells = self.visibleCells
+        let tableHeight = self.bounds.size.height
+        let tableWidth = self.bounds.size.width
+        var delayCounter = delay!
+        
+        for cell in cells { cell.transform = CGAffineTransform(translationX: tableWidth * CGFloat(directionX!), y: tableHeight * CGFloat(directionY!)) }
+
+        for cell in cells {
+            UIView.animate(withDuration: duration, delay: delayCounter * delayMultiplier!, usingSpringWithDamping: 0.8, initialSpringVelocity: 0, options: .curveEaseInOut, animations: {
```
DiegoQuiros94: if I understand this correctly you are calculating for each cell when it should start it's animation so it starts after the last cell, I think it would be better if you took advantage of the completion parameter of the animation, that way you can make each cell call the next one when it finishes animating, something like:

```
private func animateCellsFromIndex(_ index: Int ) {
           guard let cell = visibleCells[index] else { return }
            UIView.animate(withDuration: duration, delay: delayCounter * delayMultiplier!, usingSpringWithDamping: 0.8, initialSpringVelocity: 0, options: .curveEaseInOut, 
                   animations: {
                                 cell.transform = CGAffineTransform.identity
                   }, completion: { self.animateCellsFromIndex(index +1) }
}
```

then you just call it with index 0 in your function:

`animateCellsFromIndex(0)`

WolmoCore/Extensions/Animations/UIViewExtension.swift
```diff
@@ -82,3 +82,36 @@ public extension UIView {
...
+                cell.transform = CGAffineTransform.identity
+            }, completion: nil)
+            print(delayCounter * delayMultiplier!)
+            delayCounter += 1
+        }
+    }
+    
+}
```
DiegoQuiros94: what happens if the user scrolls the table while we are playing the animation? should we disable the scroll while it's playing 🤔?

DiegoQuiros94:
Looks really nice and it's very simple to use, i like it 🥇

WolmoCore/Extensions/Animations/UIViewExtension.swift
DiegoQuiros94: now that i look at the gif preview i see that there are multiple cells playing at the same time so the animations start before the previoius one completes, so this won't work. I would change the delay description, it made me understand that it was the time between the end of one animation and the start of the next one.

DiegoQuiros94: maybe change it's name to speed instead of delay, seems more representative of what it does

DiegoQuiros94:
you can see a part of the last cell on the bottom of the view that is not being animated, can we fix that?

WolmoCore/Extensions/Animations/UIViewExtension.swift
```diff
@@ -82,3 +82,36 @@ public extension UIView {
     }
     
 }
+
+public extension UITableView {
+    
+    /**
+     Adds a custom animation to the table that executes the closure when long pressed
+     
+     - Parameter duration: Time in seconds the animation will be executed for cell. Default is 1.75
```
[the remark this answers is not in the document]
Nick-m96: We can put 1.5 seconds as defult but, If the time if lower than 1.5 seconds the animation look a little forced

WolmoCore/Extensions/Animations/UIViewExtension.swift
gmazzei: Sounds good! Let's use `forEach` then.

woloxci:

:white_check_mark: Code Coverage (Slather): 84.3


Click [here](http://codestats.wolox.com.ar/organizations/wolox/projects/wolmo-core-ios/branches/animation-tableview) for more details.

PR #136, opened 2020-02-11
Add extensions for accessibility, uikit and autolayout

SalvadorCarnelutti:
Changes that Gaston wanted to add to WolmoCore before leaving. There are some constraint related methods, some convenience initializers for some UIKit element and the `applyAccessibility` method that was applied throughout all the **StayQuit** project.


NickMano (approved):

jgordowolox:
Updated documentation on-branch in case this gets merged into master before 6.4.0.

PR #142, opened 2021-11-05
Carthage to pods

jgordowolox:
## Summary ##

Removed testing libraries and re-added them vía CocoaPods


Podfile
```diff
@@ -0,0 +1,36 @@
+# Uncomment the next line to define a global platform for your project
+# platform :ios, '9.0'
```
NickMano, line 2: Uncomment this and update to ios 12

NickMano (approved):

PR #151, opened 2021-12-03
refactor: border methods

NickMano:
## Summary ##
- Refactor border extension to reduce repeated code
- Add clipToBoundsBordersIfNeeded method
- Add private setBorderFrame method

### DIFF
--- a/WolmoCore/Extensions/UIKit/UIView/Borders.swift
+++ b/WolmoCore/Extensions/UIKit/UIView/Borders.swift
@@ -20,6 +20,13 @@ public struct BorderViewProperties {
         self.color = color
         self.roundedCorners = rounded
     }
+    
+    func clipToBoundsBordersIfNeeded(borderView: inout BorderView) {
+        if self.roundedCorners {
+            borderView.layer.cornerRadius = CGFloat(self.thickness / 2)
+            borderView.clipsToBounds = true
+        }
+    }
 }
 
 /**
@@ -105,9 +112,10 @@ public extension UIView {
      */
     @discardableResult
     func add(top border: BorderViewProperties,
-             withLeftOffset left: CGFloat = 0, rightOffset right: CGFloat = 0,
+             withLeftOffset left: CGFloat = 0,
+             rightOffset right: CGFloat = 0,
              layout: LayoutMode = .constraints) -> BorderView {
-        let borderView: BorderView
+        var borderView: BorderView
         
         switch layout {
         case .constraints:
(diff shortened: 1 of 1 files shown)


WolmoCore/Extensions/UIKit/UIView/Borders.swift
```diff
@@ -249,20 +240,41 @@ public extension UIView {
...
         
-        if border.roundedCorners {
-            borderView.layer.cornerRadius = CGFloat(border.thickness/2)
-            borderView.clipsToBounds = true
+        border.clipToBoundsBordersIfNeeded(borderView: &borderView)
+        
+        return borderView
+    }
+    
+    private func setBorderFrame(border: BorderViewProperties,
```
DiegoQuiros94, line 254: setBorderFrame sounds like you are modifying the object, i would call it getBorderFrame

DiegoQuiros94: or maybe getBorderView, since it's returning a BorderView

WolmoCore/Extensions/UIKit/UIView/Borders.swift
```diff
@@ -116,19 +124,13 @@ public extension UIView {
...
+                                        firstOffset: left,
+                                        secondOffset: right,
+                                        position: .top)
         }
         
-        if border.roundedCorners {
-            borderView.layer.cornerRadius = CGFloat(border.thickness/2)
-            borderView.clipsToBounds = true
-        }
+        border.clipToBoundsBordersIfNeeded(borderView: &borderView)
```
DiegoQuiros94, line 133: I think it would be better if BorderView had the function, so you can call it like this:
`borderView.clipToBoundsBordersIfNeeded(border: border)`
Its weird that the border is modifying the view

DiegoQuiros94 (approved):
Nice 💯

WolmoCore/Extensions/UIKit/UIView/Borders.swift
NickMano: Makes sense

WolmoCore/Extensions/UIKit/UIView/Borders.swift
NickMano: Maybe createBorderView?

JuanJSilva (approved):

jgordowolox (approved):

WolmoCore/Extensions/UIKit/UIView/Borders.swift
DiegoQuiros94: sounds good too

PR #152, opened 2021-12-20
Create example view with Border and Gradient methods

NickMano:
## Summary ##
- Add nib view to sync xib and swift view
- Add Scroll Stack view controller to create a complex view with components
- Add view with preview of border methods and gradients
- Add snapshot to new view

## Screenshots ##

https://user-images.githubusercontent.com/11547820/146829107-90ae4dda-5b70-4ad3-b682-9b3d7652ab4e.mov


jgordowolox (approved):

WolmoCore/Extensions/UIKit/UIView/Borders.swift
```diff
@@ -20,6 +20,13 @@ public struct BorderViewProperties {
         self.color = color
         self.roundedCorners = rounded
     }
+    
+    func clipToBoundsBordersIfNeeded(borderView: inout BorderView) {
```
gmazzei-wolox, line 24: Is `BorderView` a struct or a class? If it is a class, I think it doesn't need `inout`.

WolmoCoreDemo/PrincipalView/View.swift
```diff
@@ -9,7 +9,7 @@
 import UIKit
 import WolmoCore
 
-final internal class View: UIView, NibLoadable {
+final internal class View: NibView {
```
gmazzei-wolox, line 12: There's already a protocol called `View` in SwiftUI. Perhaps it would be a good idea to rename this class so the types don't get in conflict.

gmazzei-wolox (approved):
Nice!

WolmoCore/Extensions/UIKit/UIView/Borders.swift
NickMano: I will check it 🐱

NickMano: That method is deprecated, I made a change in a previous PR but I forgot to delete it 🤦

WolmoCoreDemo/PrincipalView/View.swift
NickMano: I will change the name

NickMano: Ready

PR #169, opened 2022-09-13
Test fixing for SPM support

jgordowolox:
Fixing tests in a different branch. **TODO:**

- `DateSpec` date to string with format test fails even though the result is the expected outcome.
- NIB loading in general appears to be broken (seemingly due to `Bundle` errors). Tests failing in `BundleSpec`, `NibLoadableSpec`, `UICollectionViewSpec`, `UITableViewSpec`.


jgordowolox:
About `DateSpec` failing test: apparently, since `Date`s are stored as `float`, there is a fractional second difference when converting `Date` to `String` and back, so that could be the reason why it's failing. [More info here.](https://stackoverflow.com/a/59867254)

PR #170, opened 2022-12-13
Remove fastlane

NickMano:
## Summary ##
Remove Fastlane because isn't being used 
Update ruby version to 2.6 to remove the warning on Bitrise.
Remove 3 warnings related to SwiftLint


danielcharry (approved):
👌

